PR #15, opened 2024-07-11
implement alias key support

v9n:
Implement alias key concept as following:

1. create alias key: allow to generate or import a private key, and encyrpt with the same ecdsa password of operator key
2. declare alias key for the operator: bind the alias public key address to the operator (ecdsa public key) hold in AP Contract

Later on, whenever we signed, we can look up the APCOnfig contract

The operator is implemented to also check and make sure there is an alias <-> operator binding existed if the ecdsa singer key doesn't match operator address.


chrisli30 (approved):
Looks good to me!

PR #19, opened 2024-07-16
fix link to etherscan/eigenlayer on telemetry

v9n:


core/config/envs.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+var (
+	MainnetChainID  = big.NewInt(1)
+	CurrentChainEnv = ChainEnv("ethereum")
+)
+
+func IsMainnet() bool {
+	return CurrentChainEnv == HoleskyEnv
+}
```
chrisli30, line 19: I would build a map here for scalability, returning the URL based on the `ChainEnv` value. 

In a few months we would have more chains to support, so the function name `IsMainnet` will become ambiguous.

chrisli30 (approved):
LGTM, I have left a comment.

PR #20, opened 2024-07-16
Add metrics to operator

v9n:
Fixed ENG-844

Update the endpoint that generate the prometheus metric.

Convert all the metrics to vector to supports:

- operator address: This will allow us to later scrape and group by operator address.
- version: the version. Useful to find operator that has not update.  Also being used on our telemetry page

(Version is just a number for demo purpose on this screenshot to show how it's appear)
CleanShot 2024-07-23 at 04 59 40@2x

Metrics generator:

- ping check duration|total, worker looper count
- Add "task received" metrics. Right now this isn't connected with real tasks yet, but the mechanism allow us to generate metrics, and therefor can update our dashboard.
- Add retrying in the sync loop so if the grpc stream is broken, we can re-connect to fetch new tasks

Change to connection pool:

- Record versions and metric port. We need the metrics port to later on connect the prometheus endpoint to our server to scrape all operator metrics, group by operator address.

CleanShot 2024-07-23 at 05 04 00@2x

### DIFF
--- a/aggregator/pool.go
+++ b/aggregator/pool.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
+	"google.golang.org/grpc/peer"
 	timestamppb "google.golang.org/protobuf/types/known/timestamppb"
 
 	"github.com/AvaProtocol/ap-avs/core/config"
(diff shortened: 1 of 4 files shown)


chrisli30 (approved):
Looks good!

PR #28, opened 2024-10-22
Support complex task

v9n:
Initial structure to support complex task with nodes data structure. Instead of a single elemenet to be run, our task now is a chain of nodes. Each node will be evaluated. the result can be passed down for further node to consume.

This PR is just to support the task structure for now.


chrisli30 (approved):
The code looks good to me

PR #31, opened 2024-11-06
Support remote signer

v9n:
This PR allows one to de-couple the BLS key from the AVS. The BLS key is only need for intial s etup. Together with alias key, this will enable us to completely de-copuple root key from AVS.

https://github.com/Layr-Labs/cerberus
https://github.com/Layr-Labs/cerberus-api

Ref Eigenda impl - https://github.com/Layr-Labs/eigenda/pull/745


chrisli30 (approved):
LGTM

PR #32, opened 2024-11-08
Return status and error code

v9n:
Use right status and error code


chrisli30 (approved):
Looks like an ancestor of #34

PR #34, opened 2024-11-11
New task management

v9n:


aggregator/rpc_server.go
```diff
@@ -53,24 +63,22 @@ func (r *RpcServer) GetNonce(ctx context.Context, payload *avsproto.NonceRequest
...
-		SmartAccountAddress: sender.String(),
+		Wallets: wallets,
 	}, nil
 }
 
 func (r *RpcServer) CancelTask(ctx context.Context, taskID *avsproto.UUID) (*wrapperspb.BoolValue, error) {
 	user, err := r.verifyAuth(ctx)
 	if err != nil {
-		return nil, err
+		return nil, status.Errorf(codes.Unauthenticated, "invalid authentication key")
```
chrisli30, line 81: Let’s capitalize the first word of error messages, "Invalid".

Also, I’d suggest us to create a list of constant for the error message, for code re-use purpose.

chrisli30 (approved):
The code looks good. I had a minor comment. Also, I think the next step is to test the .proto file from ava-sdk-js.

aggregator/rpc_server.go
chrisli30: hmm, is this PR supposed to be on top of #33?

PR #40, opened 2024-11-22
Added API docs generation instructions to README.md

chrisli30:
I fixed a few things for our website docs today, and tried to update the API Reference.

I tried a few options to generate API references directly from .proto file, and this default method works the most smooth, so I’m adding it here. **The generated HTML version looks nice**, and markdown is okay.

My goal is to find an automatic way to update docs from the .proto file so we can rapidly update the API reference on our website /docs.

Although the documentation can now be generated, I haven’t added it to the website yet. This is because many input fields still require detailed explanations, and we need to organize the definitions by command groups to create a more user-friendly structure. I believe updating the .proto file and leveraging the protoc-gen-doc plugin can help achieve this, but I suggest we focus on testing the APIs in the Studio first before finalizing the structure.

Generated HTML screenshot
CleanShot 2024-11-21 at 23 33 46@2x

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,36 @@ foundryup
 go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
 ```
 
+### Generate API Docs from Proto Files
+1. Install the `protoc-gen-doc` plugin for `protoc`.
+   Install via Go:
+   ```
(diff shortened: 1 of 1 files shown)


v9n (commented):
looks good.

PR #53, opened 2024-12-08
extract exection log out

v9n:


protobuf/node.proto
```diff
@@ -0,0 +1,85 @@
+syntax = "proto3";
```
v9n, line 1: Split node defininiton into its own proto file . This way for node update there won't be any change to the SDK

chrisli30 (approved):
Looks good to me. 

One question is that, how should we generate executions data in sdk tests? We can set the block interval to 1 and wait for a few blocks to get the executions result?

chrisli30:
Hi @v9n, is it possible to get the total number of ListTasks? The `cursor` is a new trend of getting list items, but we do have the requirement to get and display the total number directly.

chrisli30:
Another thought: I think it would be better to standardize the response format of both calls. Personally, I prefer the format of getExecutions, which is: 
```
{
    executions: [], // this is the result
    cursor: string
}
```

v9n:
> Hi @v9n, is it possible to get the total number of ListTasks? The cursor is a new trend of getting list items, but we do have the requirement to get and display the total number directly.

yes, we can return the total in this as well.

> Another thought: I think it would be better to standardize the response format of both calls. Personally, I prefer the format of getExecutions, which is:

did you check the protocolbuf, they had standarize to the same

protobuf/avs.proto
```diff
@@ -360,10 +307,48 @@ message ListWalletResp {
...
+    int64 max_execution = 8;
+    // return how many time this task has run
+    int64 total_execution = 9;
+    int64 last_ran_at = 10;
+
+    TaskStatus status  = 11;
+    TaskTrigger trigger = 12;
+  }
+
+  repeated Item tasks = 1;
```
v9n, line 339: @chrisli30 this is how the list tasks .

```
{
  tasks: [],
  cursor
}
```

PR #55, opened 2024-12-10
fix missing max_execution parma when creating task

v9n:
I forgot to include this param so it always 0

### DIFF
--- a/model/task.go
+++ b/model/task.go
@@ -56,12 +56,13 @@ func NewTaskFromProtobuf(user *User, body *avsproto.CreateTaskReq) (*Task, error
 			Owner:              owner.Hex(),
 			SmartWalletAddress: aaAddress.Hex(),
 
-			Trigger:   body.Trigger,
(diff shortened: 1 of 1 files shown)


model/task.go
```diff
@@ -56,12 +56,13 @@ func NewTaskFromProtobuf(user *User, body *avsproto.CreateTaskReq) (*Task, error
...
-			Memo:      body.Memo,
-			ExpiredAt: body.ExpiredAt,
-			StartAt:   body.StartAt,
+			Trigger:      body.Trigger,
+			Nodes:        body.Nodes,
+			Edges:        body.Edges,
+			Memo:         body.Memo,
+			ExpiredAt:    body.ExpiredAt,
+			StartAt:      body.StartAt,
+			MaxExecution: body.MaxExecution,
```
v9n, line 65: @chrisli30 this will fix the MaxExecution always returning 0

Issue #54, opened 2024-12-09
Each trigger_type should be able to manually triggered by user


chrisli30:
This will solve a few problem:
1. end to end Simulation
2. gerenate execution immediatly with manual trigger
3. each trigger type can be triggered on demand.

```
 oneof trigger_type {
    bool                   manual  = 2;

    // run at a specific epoch, name inspired by unix `at` utility
    FixedEpochCondition    fixed_time   = 3;

    // interval such as every hour/day/ etc can be converted to cronsyntax by the sdk/studio
    CronCondition          cron    = 4;
   
    // currently the only support syntax is every <x> blocks
    BlockCondition         block   = 5;

    // support filter by event expression such as topic0, topic1, topoc2 and event_data and contract_address
    EventCondition         event   = 6;
  }
```

v9n:
Implemented in #58

Issue #52, opened 2024-12-08
Nit pick: Rename graphql_data_query to match the styles of other Node variable


chrisli30:
Since the NodeType is `GraphQLQueryNode`, I suggest us to rename the `graphql_data_query` to `graphql_query`, since other variables all match their NodeType, for example `RestAPINode rest_api`.

```
    ContractReadNode contract_read = 12;
    // Make call to a graphql endpoint
    GraphQLQueryNode  graphql_data_query = 13;
    // Make call to a HTTP endpoint
    RestAPINode       rest_api = 14;
```

v9n:
Fixed in #58

PR #58, opened 2024-12-11
Continue impl for other node type execution, support  triggering

### LINKED ISSUE #51: Nit pick: rename FixedEpochCondition to be more comprehensive
Regarding,
```
message FixedEpochCondition {
  repeated int64 epochs = 1;
}
```
I suggest us to rename the struct name of `FixedEpochCondition` to `FixedTimeCondition`, due to the below reasons:
1. FixedTime is more general and straight-forward.
2. `epochs` is already in its property and very clear, so there’s no need to do write another Epoch before it. 
3. Just in case we put some other time related properties in it. `Time` would be more general.

v9n:
Fixes #51 #52 #54

Continue adding functionality including:

- Add graphql runner
- Add contract read runner
- Add block trigger (finally)
- Support manually trigger a task.
- Set tax completed when expired or reaching max execution


chrisli30 (approved):
The code looks straight-forward 👍

PR #68, opened 2024-12-12
rename TriggerMark -> TriggerMetadata

### LINKED ISSUE #67: The TriggerMark variable name does not sound very straight-forward
Initially I could not understand what is a trigger mark, but reading this line "// TriggerMark is used to populate the `trigger.data` variable." I think we could call it either TriggerMetadata, or TriggerContext. I personally prefer the `TriggerMetadata`.

For example, the updated comments and variable name will look like:
```
// TriggerMetadata represents a lookup we can use to identify when, where, and how a trigger is activated.
// In a block trigger, we will have block_number.
// In an event trigger, we will have the log_index, tx_hash, and block_number.
// In a time-based trigger (cron/fixed time), we will have the epoch.
//
// TriggerMetadata is used to populate the `trigger.data` variable. For example, if this is a transfer
// event, you will have access to trigger1.data.from_address and trigger1.data.value.
message TriggerMetadata {
  uint64 block_number = 1;
  uint64 log_index    = 2;
  string tx_hash      = 3;
  uint64 epoch        = 4;
}
```

v9n:
Fixed #67 #69

We need to fix this sooner than later before we familiar with the name


chrisli30 (approved):
Looks good, thanks!

PR #70, opened 2024-12-12
Passing result to subsequent steps

v9n:


operator/operator.go
```diff
@@ -297,7 +297,8 @@ func NewOperatorFromConfig(c OperatorConfig) (*Operator, error) {
 	)
 	if err != nil {
 		logger.Error("Cannot create AvsWriter", "err", err)
-		return nil, err
+		// EigenLayer has update the contract and we cannot fetch the slasher anymore, we should upgrade the EigenSDK, right now we don't use it so it's ok to ignore this error
```
chrisli30, line 300: Add prefix `// TODO: ` so it’s easy to find all pending code.

chrisli30 (approved):
Approved! but I have left a few comments for review.

chrisli30 (changes requested):
Just saw the description "Data from previous step will be feeded to next step using the name as variable. if name has space or invalid character, it will be replace with _"

I’d **avoid manipulating the id of the node** cause that’s error-prone. There’s no reason that a node id is not fixed nor deterministic. Could we use ulid as the id of the node, for referencing previous nodes, to avoid and string modification?

We are good to merge this PR and iterate on this previous step identifier function.

chrisli30 (approved):

PR #85, opened 2025-01-02
Change trigger response

v9n:
Change response of trigger workflow method as below

- always return `execution id` regardless we run inline or async.
- Add a new field Return the status: queue|completed  so the caller can know whether the call succeed or not

Add a new endpoint

- getExecutionStatus return. the status of an execution, ideally for caller to poll whether an execution has finished or not.

Once the execution is run, this same `execution id` will also appear on the `listExecutions`


chrisli30 (approved):
Yep, these behavior align with our design.

PR #101, opened 2025-01-15
Add secrets

### LINKED ISSUE #98: Create secrets.telegram_notify_bot_token so we can use it in the URL of RestApi
Although secrets all start with `secrets.`, they have different scopes. We don’t need to implement the permission restriction for the scope now, but let’s make sure secrets have such attribute `scope`.

My understanding is that each secret object should contain the below properties, and we can use an array to store them [secretObj, secretObj, secretObj]

name,    // the variable’s name
value,    // the variable’s value
scope:
- global    // Consider calling this "app", as application-level, for example, telegram notification is just an application from our program.
- org    // Organization level, it’s a place-holder for Github organization level access
- project    // We could optionally call it workflow, but project sounds more general, and in future we might create project which can contain multiple workflows, similar to a Github repo

v9n:
Fix #98

### DIFF
--- a/aggregator/task_engine.go
+++ b/aggregator/task_engine.go
@@ -25,7 +25,8 @@ func (agg *Aggregator) startTaskEngine(ctx context.Context) {
 	})
 	agg.worker = apqueue.NewWorker(agg.queue, agg.db)
 	taskExecutor := taskengine.NewExecutor(agg.db, agg.logger)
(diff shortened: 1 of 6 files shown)


chrisli30 (commented):
Is the `scope` mentioned in the #98 implemented in this PR?

PR #102, opened 2025-01-15
Update authentication keys

### LINKED ISSUE #97: Update authKey signature text to be more comprehensive
This is the text I saw from other apps, which looks more comprehensive and production-fit. Let’s use this format to improve our signature message for getting the authKey?

```
Please sign the below text for ownership verification.

URI: https://app.avaprotocol.org
Chain ID: 1
Version: 1
Nonce: 9d67fb24cb9a8bccffc2af2206a93b75817d8db2ef2df860442c9e97c47a908d
Issued At: 2025-01-08T19:56:17.141Z
Expire At: {expired_at_based_on_issued_at}
```

A few things to notice:
1. Among the field, I’m not sure what Nonce is used for. Since this nonce is not an integer, it’s not the same of the transaction nonce of the wallet. We can not include Nonce for now. 
2.`Version` is optional, if we don’t want to keep a version string in our program.
3. Maybe we should include the wallet address to these field too, just to make sure it’s the connected wallet signing this message.

v9n:
Fix #97

### DIFF
--- a/aggregator/aggregator.go
+++ b/aggregator/aggregator.go
@@ -51,12 +51,12 @@ const (
 func RunWithConfig(configPath string) error {
 	nodeConfig, err := config.NewConfig(configPath)
 	if err != nil {
-		panic(fmt.Errorf("failed to parse config file: %s\nMake sure it is exist and a valid yaml file %w.", configPath, err))
+		panic(fmt.Errorf("Failed to parse config file: %s\nMake sure it is exist and a valid yaml file %w.", configPath, err))
 	}
 
 	aggregator, err := NewAggregator(nodeConfig)
 	if err != nil {
-		panic(fmt.Errorf("cannot initialize aggregrator from config: %w", err))
+		panic(fmt.Errorf("Cannot initialize aggregrator from config: %w", err))
 	}
 
 	return aggregator.Start(context.Background())


aggregator/aggregator.go
```diff
@@ -51,7 +51,7 @@ const (
 func RunWithConfig(configPath string) error {
 	nodeConfig, err := config.NewConfig(configPath)
 	if err != nil {
-		panic(fmt.Errorf("failed to parse config file: %w\nMake sure %s is exist and a valid yaml file %w.", configPath, err))
+		panic(fmt.Errorf("failed to parse config file: %s\nMake sure it is exist and a valid yaml file %w.", configPath, err))
```
chrisli30, line 54: Let’s fix the capitalization now - the first letter of a sentence must be capitalized. Please fix other instances of fmt.Errorf() too. 

Use `Failed to`, not `failed to`.

Do the below return return the message string to the gRPC client?
```
return fmt.Errorf("failed to parse config file: %s\nMake sure it is exist and a valid yaml file %w.", configPath, err)
```

chrisli30 (commented):
Looks good but I left a comment.

chrisli30 (approved):

PR #110, opened 2025-01-23
disallow manual trigger of completed tasks

### LINKED ISSUE #92: manually trigger a completed workflow should failed
if a workflow is done, manually trigger it should failed.

if a workflow is completed, any queue run shoud also failed to run

v9n:
Fixes #92

### DIFF
--- a/model/task.go
+++ b/model/task.go
@@ -133,7 +133,7 @@ func (t *Task) OwnedBy(address common.Address) bool {
 // A task is runable when both of these condition are matched
 //  1. Its max execution has not reached
 //  2. Its expiration time has not reached
-func (t *Task) Runable() bool {
+func (t *Task) IsRunable() bool {
 	// When MaxExecution is 0, it is unlimited run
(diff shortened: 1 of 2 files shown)


model/task.go
```diff
@@ -130,6 +130,18 @@ func (t *Task) OwnedBy(address common.Address) bool {
 	return strings.EqualFold(t.Owner, address.Hex())
 }
 
+// A task is runable when both of these condition are matched
+//  1. Its max execution has not reached
+//  2. Its expiration time has not reached
+func (t *Task) Runable() bool {
```
chrisli30, line 136: I’m not sure about the naming in go, but to be consistent with the name `TaskIsNotRunable`, I’d call this function `isRunnable()`. The `is` will clearly indicate that this is boolean returning function.

chrisli30 (approved):
Looks good, I left one minor comment.

model/task.go
chrisli30: Yeah, understood and sounds good!

chrisli30 (approved):

PR #111, opened 2025-01-23
Add contract write runner

### LINKED ISSUE #86: Add contract write runner
We had the old code of original runner, just need to port to our new runner structe with step.

v9n:
Fix #86


chrisli30 (approved):
Looks good

PR #112, opened 2025-01-27
implement user define secret management api

v9n:


core/taskengine/doc.go
```diff
@@ -24,5 +24,10 @@ Storage can also be inspect with telnet:
 	telnet /tmp/ap.sock
 
 Then issue `get <ket>` or `list <prefix>` or `list *` to inspect current keys in the storage.
+
+**Secret Storage**
+- currently org_id will always be _ because we haven't implemented it yet
+- workflow_id will also be _ so we can search be prefix
```
chrisli30, line 30: I don’t quite get this line, `- workflow_id will also be _ so we can search be prefix`

Does this mean the search is not done at the workflow level, but only user level? I’m fine with that, but let’s clarify this comment.

core/taskengine/engine_test.go
```diff
@@ -453,5 +453,323 @@ func TestTriggerCompletedTaskReturnError(t *testing.T) {
 	if err == nil || resultTrigger != nil {
 		t.Errorf("expect trigger error but succeed")
 	}
+}
+
+func TestCreateSecret(t *testing.T) {
```
chrisli30, line 458: Currently we don’t need to add the same tests in sdk, if unit tests are done here, but I don’t have good visibility what are covered in the unit tests. 

For this module, let’s rely on these units for testing, but simple tests are still needed in the sdk, so I can look up the usage of the secrets sdk code.

chrisli30 (approved):
Looks good!

core/taskengine/doc.go
chrisli30: oh I see. Makes sense 👌

PR #121, opened 2025-02-04
Decode the method name from calldata automatically

### LINKED ISSUE #120: Check out method in SmartContractWrite and see if we remove the redundency
The method is already inside the calldata, let’s have a single source of truth for method. If method is required inside AVS, parse it out from the API. 

https://github.com/AvaProtocol/ava-sdk-js/blob/0322812bc501c5e4fc8c80e29d7ad42a9aecb6a8/grpc_codegen/avs.proto#L123

v9n:
Fix #120


chrisli30 (approved):
Looks good!

PR #125, opened 2025-02-07
Do not create default wallet

v9n:
We now require explitcitly create a wallet before use.


chrisli30 (approved):
Tested this branch with Studio. It looks good 👍

PR #138, opened 2025-02-14
add a pre processing text with js evaluation for `{{ }}` block

### LINKED ISSUE #133: Change how we template the variable and secret
Follow retool.

Use `{{  name }}` where name can be anything such as:

- nodename.data: data from previous. node
- apContext.configVars.<secret_name>: access any secret 

Also, the syntax `{{ expression }}` can also be used in place like rest body. And the expression can be though of a single JavaScript expression. Such as we can compose a restbody like this

```
my body is {{ JSON.stringify(someNodeName.data) }} 
```

In JavaScript code, we don't need `{{ }}` and can just access `apContext` directly.

v9n:
Fix #133 

This PR does a few things:

- Support using secrets in custom code. It was only in Rest node before.
- Use a new syntax for both. of variable access + secret access describe in #133 similar to retool.

You now can write these kind of body:

```
This is my code {{ JSON.stringify(dataFromPreviousNode.data) }} also {{ dataFromPreviousNode.data.value }}
```

## How this work

For code block, these vars/secrets(apContext) or `nodeName` are normal variable and can access the data inside them with `.data` or `apContext.configVars.secretname`

For non code block, such as Rest or GraphqQL, whenever we need to serelized a piece of text to send over the wire (URL, header, body), we will run a pre-processing step on that text. All of `{{ }}` block is evaluate as JavaScript. So inside it you can use normal primitive of JavaScript.


chrisli30 (approved):
LGTM! @v9n, after merging this change, could you also place the variable into this line of the secret test, to see if it could read it? (maybe we should use a REST API node, since the value can’t be printed in log)

chrisli30 (approved):
The ava-sdk secret tests are passing with this PR change. Approved!

PR #155, opened 2025-02-19
add account abstraction related contract address

v9n:
We use a consistent contract address across four networks:

- **Base Sepolia**  
- **Sepolia**  
- **Base**  
- **Ethereum**  

Addresses are added


chrisli30 (approved):
nice, all 4 addresses look the same!

PR #166, opened 2025-02-21
Rename raw log and event

v9n:
Rename a couple field in OutputData accordingly to this slack discussion.

https://avaprotocol.slack.com/archives/D0570FNE0SJ/p1740162219146389

Address 2 points

- okay, so it is the log of an Evm chain. Could we call it EvmLog, instead of rawEvent?
- To be consistent, for the transfer, let’s call it transferLog too, so both log are on-chain related.


chrisli30 (approved):
Looks good!

PR #182, opened 2025-03-10
require opt-in to enable event monitoring

v9n:
Event monitoring can be costly depend on RPC provider so we require an opt-in to use it.

### DIFF
--- a/config/operator_sample.yaml
+++ b/config/operator_sample.yaml
@@ -37,6 +37,6 @@ target_chain:
   eth_rpc_url: <sepolia-chain-rpc>
   eth_ws_url: <websocket-sepolia-chain-rpc>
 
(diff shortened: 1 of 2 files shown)


config/operator_sample.yaml
```diff
@@ -36,3 +36,7 @@ db_path: /tmp/ap-avs-operator
 target_chain:
   eth_rpc_url: <sepolia-chain-rpc>
   eth_ws_url: <websocket-sepolia-chain-rpc>
+
+enable_features:
```
chrisli30, line 40: `enabled_features`, to match the variable `EnabledFeatures` below.

chrisli30 (approved):
Looks good, but I have one minor comment.

PR #207, opened 2025-04-10
Define migration process and perform the first migration for second to ms

v9n:
This PR implements:

- Auto format code on push and make commit back. No need to run any formatter locally
- Define a migration framework. The migration will perform a backup before run. Backup can be taken through repl.
- Define our first ever migration that convert timestamp in second to ms.

Note that BadgerDB is rsync-friendly and we will have a different backup with rsync on cron on our aggregator use `rsync`


chrisli30 (approved):
Looks good!

PR #236, opened 2025-05-12
Adde Slasher method interface to AutomationServiceManager

chrisli30:


chrisli30:
The operators are running without needing this PR. Close for now.

PR #534, opened 2026-05-29
fix: parseABIParameter dispatches on UintTy/IntTy size

chrisli30:
## Summary

Fixes the `abi: cannot use ptr as type uint16 as argument` failure that surfaced when running the v4 AAVE template tests against the Railway gateway (see [ava-sdk-js#213](https://github.com/AvaProtocol/ava-sdk-js/pull/213)). The bug had been dormant in the canonical AAVE template because its branch never fired in dev; with the branch threshold corrected and the dev wallet now carrying an AAVE position, every supply call crashed here.

## Root cause

[`parseABIParameter`](core/taskengine/utils.go) returned `*big.Int` for **every** uint/int ABI type, regardless of width:

```go
case abi.UintTy, abi.IntTy:
    return bigint.Parse(param)  // always *big.Int
```

go-ethereum's ABI encoder is type-strict: it expects native Go `uint8`/`uint16`/`uint32`/`uint64` for fixed-size widths ≤ 64 bits and `*big.Int` only for the larger widths (uint128, uint256, etc.). Passing `*big.Int` for `uint16` raises:

```
abi: cannot use ptr as type uint16 as argument
```

The "ptr" in the error is the `*big.Int` pointer; the encoder expects a value of the native type.

End-to-end repro from the failing run:

```
[ERRO] ❌ Failed to get/generate calldata for contract write
       methodName="supply"
       rawMethodParams=[..."0xedC0945A6e3AC235AfBCb408b0638117C3ba2940","0"]
       error="failed to pack method call: abi: cannot use ptr as type uint16 as argument"
       caller="taskengine/vm_runner_contract_write.go:323"
```

The `referralCode = "0"` (the last methodParam — a `uint16`) parses fine but lands in `abi.Pack` as `*big.Int`.

## Fix

[`parseABIParameter`](core/taskengine/utils.go) now dispatches on `abi.Type.Size` for `UintTy` and `IntTy`:

- Size ≤ 64: down-convert `*big.Int` to the native Go type (`uint8`/`uint16`/`uint32`/`uint64`, `int8`/`int16`/`int32`/`int64`) with explicit overflow + sign validation.
- Size > 64: pass through as `*big.Int` (covers `uint128`/`uint256`/`int128`/`int256`).
- Range errors carry the offending value + width so future regressions of this shape surface with a clear message instead of a panic deep in go-ethereum.

## Tests

Three new test functions, 21 sub-cases, all pass (`go test ./core/taskengine -run "TestParseABIParameter|TestGenerateCallData_AAVE"`):

- **`TestParseABIParameter_FixedSizeIntegerCoercion`** pins the dispatch for every fixed-size uint/int width (8, 16, 32, 64) plus uint256 and int256 boundary checks. Uses `require.IsType` so a regression that re-returns `*big.Int` for uint16 would fail immediately.
- **`TestParseABIParameter_FixedSizeOverflow`** pins the overflow + sign validation: `uint16 > 65535`, `uint8 > 255`, negative for any uint, `int16 > 32767`, `int16 < -32768`.
- **`TestGenerateCallData_AAVESupplyReferralCode`** is the end-to-end repro using AAVE V3 Pool.supply's actual ABI (`address asset, uint256 amount, address onBehalfOf, uint16 referralCode`) with the exact methodParams shape the v4 SDK template test sends. A regression of the original bug would fail this test directly.

## Branch base

This PR targets `main`, not `feature/rest-api-migration`. The fix touches only `core/taskengine/utils.go` and a test file, has no dependency on REST-migration work, and `feature/rest-api-migration` has an unrelated build break (functions `chainIDFromExecutionContext` and `workflowChainID` are referenced in `summarizer_context_memory.go` but never defined anywhere in the tree — separate issue, already has WIP fixes in someone's working tree). Once this lands on `main`, merging `main → feature/rest-api-migration` picks it up for the gateway deploy.

## Test plan

- [x] `go build ./...` clean (the `core/taskengine` package; scripts/main duplicates are a pre-existing unrelated issue).
- [x] `go test ./core/taskengine -run "TestParseABIParameter|TestGenerateCallData_AAVE|TestGenerateCallData_InvalidNumericValue"` — 21/21 pass.
- [ ] Once merged + `main` is pulled into the deployed branch, re-run [ava-sdk-js#213](https://github.com/AvaProtocol/ava-sdk-js/pull/213)'s AAVE template tests against the Railway gateway. The canonical template's supply step should now succeed (or fail downstream on actual chain state, not on calldata generation).


chrisli30:
Closing — this PR mixed concerns. The ABI parseABIParameter fix is being deferred; opening a focused replacement that only addresses the summarizer_context_memory.go build break, properly based on main.

PR #542, opened 2026-06-04
chore: add merge_hetzner_into_gateway one-shot migration tool

chrisli30:
## Why

To preserve the ~3 weeks of production data that Studio wrote to the Hetzner aggregators between the May 2026 Railway migration and the 2026-07-01 Hetzner decom. Full context + per-prefix policy matrix lives in [`avs-infra/docs/changes/20260604-hetzner-data-restore-plan.md`](https://github.com/AvaProtocol/avs-infra/blob/main/docs/changes/20260604-hetzner-data-restore-plan.md).

This PR adds the tool. The user's per-prefix matrix has been signed off; the tool is built to match.

## What it does

`scripts/migration/merge_hetzner_into_gateway` is a one-shot tool run **once per donor BadgerDB**, sequentially, with the donor's chain ID. For each prefix it knows about, it applies a documented policy:

| Prefix family | Policy |
|---|---|
| `t:`, `u:`, `history:` | Pass-through chain-scoped key; validate embedded chain ID matches donor |
| `w:`, `wsalt:`, `fl:`, `fr:` | Stamp donor chain ID into the legacy chain-implicit key |
| `secret:` | Import as-is (already user/org/workflow scoped) |
| `execution_index_counter:` | Import; on collision take the max |
| `ct:cw:`, `pending:`, `trigger:`, `migration:` | Drop |

**Every write is `setIfAbsent`** — `Exist` check before every `Set`. Nothing the gateway has already written is ever overwritten. The test `TestSetIfAbsent_SkipsWhenKeyPresent` is the regression guard for this; it asserts that even when both sides have the same key with different values, the gateway's value is preserved verbatim.

Unknown prefixes hard-fail by default (`--fail-on-unknown-prefix=true`) so a future schema addition can't silently drop on the floor.

## Files

| File | What |
|---|---|
| `main.go` | Flag parsing, DB open/close, prefix iteration, unknown-prefix scan, summary printing |
| `handlers.go` | Dispatch table + per-prefix handler funcs + `setIfAbsent` + chain-ID parsing helpers |
| `stats.go` | Per-prefix stat counters + summary table formatter |
| `handlers_test.go` | 21 test cases covering every handler + dispatch routing + safety properties |
| `README.md` | Operational runbook: setup, flags, output format, safety rules, what's NOT in the tool |

## Test plan

- [x] `go build ./scripts/migration/merge_hetzner_into_gateway/...` clean
- [x] `go vet ./scripts/migration/merge_hetzner_into_gateway/...` clean
- [x] `go test -count=1 ./scripts/migration/merge_hetzner_into_gateway/...` — all 21 test cases pass
- [ ] CI passes
- [ ] After merge: run against a scratch copy of one donor in dry-run mode and confirm the per-prefix counts match a manual `restic ls`. NOT to be run against the live gateway volume until 2026-06-15+.

## Out of scope (separate work)

The tool **stamps** `w:` / `wsalt:` / `fl:` / `fr:` keys with the donor chain ID, producing chain-scoped form on the gateway. The gateway's READ paths for those families still expect chain-implicit form today. A separate PR will update the gateway code to look at chain-scoped form for these prefixes — that ships before the actual merge runs against production. Including the read-path changes in the same PR was avoided so reviewers can audit the tool independently of the gateway changes.

Also out of scope:
- `wsalt:` re-derivation. The tool stamps the existing index entries; after the merge runs, the operator should also run `scripts/migration/backfill_wallet_salt_index` against the gateway to rebuild canonical pointers if any factory upgrades happened during the donor window.
- Recovery from a botched merge. The plan is "restore the pre-merge gateway backup." No in-tool rollback.

## Why `chore:`

This is a one-shot operational tool, not a user-facing feature. `chore:` ensures `go-semantic-release` doesn't cut an unnecessary release. The tool ships whenever the next user-facing change in EigenLayer-AVS does.

### DIFF
--- a/scripts/migration/merge_hetzner_into_gateway/README.md
+++ b/scripts/migration/merge_hetzner_into_gateway/README.md
@@ -1,0 +1,133 @@
+# merge_hetzner_into_gateway
+
+One-shot tool to merge a Hetzner per-chain aggregator BadgerDB (the **donor**) into the unified Railway gateway BadgerDB. Used once during the 2026-07-01 Hetzner decom to bring forward the ~3 weeks of production data that Studio wrote to the Hetzner aggregators after the May 2026 Railway migration but before this cutover.
+
+Full context: [`avs-infra/docs/changes/20260604-hetzner-data-restore-plan.md`](https://github.com/AvaProtocol/avs-infra/blob/main/docs/changes/20260604-hetzner-data-restore-plan.md).
+
+## What it does
+
+For each prefix it knows about, the tool walks every donor key and applies a per-prefix policy:
+
+| Prefix family | Policy |
+|---|---|
+| `t:` | Pass-through (chain-scoped; validate embedded chain ID matches donor) |
+| `u:` | Pass-through (chain-scoped) |
+| `history:` | Pass-through (chain-scoped) |
+| `w:` | Stamp donor chain ID into the key (legacy → chain-scoped) |
+| `wsalt:` | Stamp donor chain ID; **post-merge, run** `BackfillWalletSaltIndex` against the gateway DB to rebuild any salt-index drift |
+| `fl:` | Stamp donor chain ID |
+| `fr:` | Stamp donor chain ID |
+| `secret:` | Import as-is (already user/org/workflow scoped, chain doesn't apply) |
+| `execution_index_counter:` | Import as-is; on collision keep the larger value |
+| `ct:cw:` | Drop (cosmetic) |
+| `pending:` | Drop (transient) |
+| `trigger:` | Drop (reconstructible from `history:`) |
+| `migration:` | Drop (donor's migration history, not gateway's) |
+
+**Every write is skip-if-exists.** The gateway has been accumulating its own data since deploy — the tool never overwrites anything already there. This is the load-bearing safety property; the test `TestSetIfAbsent_SkipsWhenKeyPresent` pins it.
+
+Any prefix the tool does not recognize causes a hard-fail by default. To inspect a problem donor without aborting, pass `--fail-on-unknown-prefix=false` — but be aware that those keys are silently dropped from the merge.
+
+## Usage
+
+```bash
+# 1. Stop the donor aggregator and capture its BadgerDB to a workstation:
+ssh ap-prod1 'sudo systemctl stop avs-aggregator-ethereum'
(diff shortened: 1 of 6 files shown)


chrisli30:
All 13 Copilot comments addressed in 2788a1f. Summary of changes:

**Correctness**
- `handleMaxOnCollision` now uses `gateway.Exist()` instead of string-matching `"Key not found"` on the `GetKey` error (was brittle to error-string changes).
- Collision-resolved overwrites now increment **both** `stat.copied` and `stat.collisionResolved`, so summary totals reconcile. `CollRes` is now a documented subset of `Copied`.
- `scanForUnknownPrefixes` no longer calls `GetByPrefix([]byte(""))` — that materialized every donor key+value into memory and was the OOM risk Copilot flagged. Replaced with a new `storage.Storage.IterateKeysOnly(prefix, visit)` method (added in [storage/db.go](https://github.com/AvaProtocol/EigenLayer-AVS/pull/542/files#diff-storagedbgo)) that uses Badger's `KeysOnly` iterator — constant memory in the size of the donor.

**Documentation**
- Top-level operational-constraints block in [main.go](https://github.com/AvaProtocol/EigenLayer-AVS/pull/542/files), the `setIfAbsent` doc comment in [handlers.go](https://github.com/AvaProtocol/EigenLayer-AVS/pull/542/files), and the README's "Every write is skip-if-exists" callout all now explicitly carve out the one `execution_index_counter:` max-on-collision exception. They were misleading before.
- Gateway-open error message now mentions that the read-write lock is acquired even for `--dry-run` (so `--dry-run` operators don't get confused when the gateway is still up).
- `scanForUnknownPrefixes` doc no longer references a nonexistent "first 32 bytes" truncation; it accurately describes the "everything up to first colon" bucketing.

**Summary table**
- New `CollRes` column in per-prefix rows and the `TOTAL` line; column widths padded consistently.

**Test hygiene**
- Removed the dead `strconv` import + `var _ = strconv.Itoa` placeholder from `handlers_test.go`.
- `TestHandleMaxOnCollision_DonorLarger` now asserts both `stat.copied == 1` AND `stat.collisionResolved == 1`, pinning the "CollRes is a subset of Copied" semantics.

All 21 tests still pass. Build + vet clean across the whole repo.

PR #545, opened 2026-06-04
feat: end-to-end Hetzner→gateway migration rehearsal tooling + merge-tool dispatch fixes

chrisli30:
## Why

End-to-end test of the Hetzner → Railway-gateway merge tool against fresh donor BadgerDB snapshots pulled from all 4 production Hetzner aggregators, run on a local dev machine. Tonight's rehearsal surfaced 4 real merge-tool issues that would have silently failed in production — all fixed in this PR. After the fixes, all 267,451 user-data keys across 4 chains merge with zero errors and the gateway can read everything.

### 1. Every Hetzner donor predates `MigrateKeysToChainScoped`

`t:`, `u:`, and `history:` keys on disk are all in LEGACY form (`t:c:taskID`, not `t:<chainID>:c:taskID`). The prior dispatch routed them through `handleChainScopedPassThrough` which refused with `"donor is in legacy (pre-chain-scoped) form — run the chain-scope migration on the donor before merging"`.

**Fix**: route `t:`/`u:`/`history:` through `handleStampChainID`, extended to handle BOTH legacy and already-chain-scoped donors. Mis-stamp protection: a donor key `t:8453:c:taskID` merged with `--donor-chain-id=1` hard-fails instead of silently producing `t:1:8453:c:taskID`. Folds the legacy-to-chain-scoped rewrite into the merge itself — we never have to mutate donor data before merging.

### 2. Two production prefixes weren't in the dispatch table

- `operator:` — per-aggregator operator pool registry (3-27 keys per donor)
- `q:` — per-aggregator async execution queue (**7-10 MILLION keys** per mainnet donor)

Both are per-aggregator operational state, irrelevant to the unified gateway. Added as `handleDrop`. Without this the merge would have hard-failed with `⚠️  Donor has keys with N prefix(es) not recognized` and we'd have had to abort the maintenance window.

### 3. Per-DB Badger sequence counters

`t:seq` (engine's task ID sequence) and `q:seq:*` (queue sequences) are per-aggregator. Added as DROP entries ahead of `t:` / `q:` in the dispatch table (order matters — first-match wins).

### 4. Stats accounting double-count

Outer-loop `GetByPrefix("t:")` returned the `t:seq` key too, then `GetByPrefix("t:seq")` returned it AGAIN — counted in both rows of the summary. Added `outerLoopPrefixes()` to skip prefixes that are strict extensions of another known prefix. Stats now reconcile cleanly.

## Tooling added

| File | What |
|---|---|
[...]

PR #544, opened 2026-06-04
feat: chain-scoped gateway storage + Hetzner→Railway migration pipeline + env-var hygiene

chrisli30:
## Why

Promote staging → main so the full Hetzner→Railway migration pipeline ships to production in one coordinated cut: chain-scoped storage, the merge tool, the end-to-end rehearsal tooling that we ran live against production donors tonight, the Railway env-var hygiene, and the operator chain-registry resilience fix.

## What's in this batch (8 commits, ~2,000 lines net)

| Commit | Type | What |
|---|---|---|
| 3f24295 | `refactor:` | [#546](https://github.com/AvaProtocol/EigenLayer-AVS/pull/546) — derive `eth_ws_url` from `eth_rpc_url` in code, rename Railway env vars `${ETHEREUM_DWELLIR_RPC}` → `${ETHEREUM_RPC}` (etc.) so RPC provider names aren't baked into env vars. **10 Railway env vars → 5.** Operator + aggregator config parsers both derive WS so removing `eth_ws_url` from operator YAML doesn't leave it empty (caught by Copilot review). |
| ccb65a1 | `feat:` | [#545](https://github.com/AvaProtocol/EigenLayer-AVS/pull/545) — end-to-end Hetzner→gateway migration rehearsal tooling + merge-tool body cleaner + engine `DiscardUnknown` defense-in-depth. Tonight's rehearsal ran the full pipeline against real production data from all 4 chains; surfaced + fixed 4 real bugs that would have silently broken the production merge. |
| f3763ed | `feat:` | [#543](https://github.com/AvaProtocol/EigenLayer-AVS/pull/543) — chain-scope `w:` / `wsalt:` / `fl:` / `fr:` storage keys end-to-end (engine + executor + vm + validation + migration tools + tests) |
| c8e004e | `chore:` | [#542](https://github.com/AvaProtocol/EigenLayer-AVS/pull/542) — `scripts/migration/merge_hetzner_into_gateway` one-shot tool + `storage.IterateKeysOnly` streaming iterator |
| 184eb7b | `fix:` | chain_registry: non-blocking gRPC dial + retry-on-error RPC client cache (gateway no longer hangs when one chain's worker is briefly down) |
| da70c07 | `docs:` | drop explicit `eth_rpc_url` / `eth_ws_url` from operator config example |
| 2c0a82f | `docs:` | fix operator setup example to use current mainnet config |
| fcdebee | `chore:` | bump version.go semver default `1.9.6` → `3.0.0` to match released tag |

[...]

PR #546, opened 2026-06-04
refactor: derive eth_ws_url + rename Railway RPC env vars

chrisli30:
## Why

Two related cleanups that fall out of tonight's rehearsal:

1. **WebSocket URL derivation in code** — every RPC provider we use (Dwellir, Tenderly, Alchemy, Infura, mainnet.base.org) serves HTTP and WS at the same host+path with only the scheme changing. Maintaining `eth_rpc_url` + `eth_ws_url` as two env vars per chain just doubled the number of secrets the operator has to rotate in lockstep — and "rotate RPC but forget WS" silently breaks WS-only code paths.
2. **Drop the `_DWELLIR_` provider name from Railway env vars** — bakes the RPC provider into the env-var name, which is wrong twice over: the gateway code doesn't care who the provider is, and we'd have to rename every var the day we move off Dwellir.

## Net change

| Before | After |
|---|---|
| 10 Railway env vars (5 chains × 2: HTTPS + WS) | **5 Railway env vars** (5 chains × 1 HTTPS, WS derived) |
| `ETHEREUM_DWELLIR_RPC` + `ETHEREUM_DWELLIR_WS` | `ETHEREUM_RPC` |
| `BASE_DWELLIR_RPC` + `BASE_DWELLIR_WS` | `BASE_RPC` |
| `SEPOLIA_DWELLIR_RPC` + `SEPOLIA_DWELLIR_WS` | `SEPOLIA_RPC` |
| `BASE_SEPOLIA_DWELLIR_RPC` + `BASE_SEPOLIA_DWELLIR_WS` | `BASE_SEPOLIA_RPC` |
| `BNB_DWELLIR_RPC` + `BNB_DWELLIR_WS` | `BNB_RPC` |

## What changes in code

[`core/config/config.go`](../blob/refactor/env-vars-and-derive-ws/core/config/config.go) — new `deriveWsURL(rpcURL string) string` helper that swaps the scheme (`https` → `wss`, `http` → `ws`). Called when the YAML's `eth_ws_url` is empty, for both the top-level config and each chain's `smart_wallet`. An explicit `eth_ws_url` still wins for the rare case where WS lives at a separate URL.

[`core/config/derive_ws_url_test.go`](../blob/refactor/env-vars-and-derive-ws/core/config/derive_ws_url_test.go) — new test covering `https://`, `http://`, empty, and unknown-scheme cases.

## What changes in YAML (8 files)

`config/gateway-railway.yaml`, `worker-*-railway.yaml` (5 files), `operator-ethereum-railway.yaml`, `operator-railway.yaml` — all `eth_ws_url:` lines removed, all `${X_DWELLIR_RPC}` references renamed to `${X_RPC}`, header comments updated.

Local dev configs (`gateway-dev.yaml`, `gateway-dev-rehearsal.yaml`, `worker-*-dev.yaml`) untouched — their explicit `eth_ws_url` values still work because the derivation only kicks in when the value is empty.

## Railway action required when this lands

Each consuming service needs the new var names set with the HTTPS URL value (the live prod Dwellir key `eb302287-...c9d` per the 2026-06-04 operator decision):

| Service | Vars to set |
|---|---|
| `gateway` | `ETHEREUM_RPC`, `BASE_RPC`, `SEPOLIA_RPC`, `BASE_SEPOLIA_RPC`, `BNB_RPC` |
| `worker-ethereum` | `ETHEREUM_RPC` |
| `worker-base` | `BASE_RPC` |
| `worker-sepolia` | `SEPOLIA_RPC` |
| `worker-base-sepolia` | `BASE_SEPOLIA_RPC` |
| `worker-bnb-mainnet` | `BNB_RPC` |
| `operator-ethereum` | `ETHEREUM_RPC` |
| any other `operator-*` | the chain RPC they use today |

Set each as `https://api-<chain>-mainnet.n.dwellir.com/eb302287-4dc1-47de-8e5a-ad0f6cc41c9d` (for production) — see the production tfstate-on-disk paths in [`avs-infra/terraform/environments/production/ethereum-aggregator.yaml`](https://github.com/AvaProtocol/avs-infra/blob/main/terraform/environments/production/ethereum-aggregator.yaml) for the exact URLs.

**After deploy**, the old `_DWELLIR_RPC` / `_DWELLIR_WS` vars can be deleted from Railway. Tracking as a separate todo.

## Test plan

- [x] `go build ./...` clean
- [x] `go test -count=1 -short ./core/config/` — passes (including new `TestDeriveWsURL`)
- [x] `grep DWELLIR config/*-railway.yaml` — no matches
- [x] `grep eth_ws_url config/*-railway.yaml` — no matches (all derived)
- [ ] CI passes
- [ ] After merge: set Railway env vars (above), deploy, verify each service's startup log shows the correct chain ID + connects to RPC

## Related

- PR #544 (staging→main promotion, includes chain-scoped storage + merge tool)
- PR #545 (rehearsal tooling + body cleaner + DiscardUnknown loader fix)
- This PR — env var hygiene, doesn't block #544/#545

### DIFF
--- a/config/gateway-railway.yaml
+++ b/config/gateway-railway.yaml
@@ -17,11 +17,15 @@
 #       variables so the topology auto-resolves and survives service
 #       renames, e.g.:
 #         WORKER_ETHEREUM = ${{worker-ethereum.RAILWAY_PRIVATE_DOMAIN}}:50051
-#   ETHEREUM_DWELLIR_RPC / ETHEREUM_DWELLIR_WS  — AVS chain RPC
-#   SEPOLIA_DWELLIR_RPC  / SEPOLIA_DWELLIR_WS   — sepolia chain RPC
-#   BASE_DWELLIR_RPC     / BASE_DWELLIR_WS
-#   BASE_SEPOLIA_DWELLIR_RPC / BASE_SEPOLIA_DWELLIR_WS
-#   BNB_DWELLIR_RPC      / BNB_DWELLIR_WS       — BNB Smart Chain mainnet
+#   ETHEREUM_RPC          — Ethereum mainnet HTTPS RPC (WS derived in code)
+#   SEPOLIA_RPC           — Sepolia HTTPS RPC
+#   BASE_RPC              — Base mainnet HTTPS RPC
+#   BASE_SEPOLIA_RPC      — Base-Sepolia HTTPS RPC
+#   BNB_RPC               — BNB Smart Chain mainnet HTTPS RPC
+# (WebSocket URLs are auto-derived from the HTTPS URL by core/config —
+# every supported provider serves both transports at the same host+path
+# with only the scheme differing. Set the explicit eth_ws_url in the
+# YAML only if a chain genuinely needs a separate WS endpoint.)
 #   ETHEREUM_BUNDLER_URL, SEPOLIA_BUNDLER_URL,
 #   BASE_BUNDLER_URL,     BASE_SEPOLIA_BUNDLER_URL
 #   BNB_BUNDLER_URL      — optional; connectivity-only rollout leaves it unset
@@ -43,8 +47,7 @@ ecdsa_private_key: ${GATEWAY_ECDSA_PRIVATE_KEY}
 # disabled, so one gateway can serve operators from any AVS chain. We
 # point this at Sepolia AVS to match the existing testnet operator;
 # flip to Ethereum mainnet AVS when enforceAuth is turned on.
-eth_rpc_url: ${SEPOLIA_DWELLIR_RPC}
-eth_ws_url: ${SEPOLIA_DWELLIR_WS}
+eth_rpc_url: ${SEPOLIA_RPC}
 
 rpc_bind_address: "0.0.0.0:2206"
 http_bind_address: "0.0.0.0:8080"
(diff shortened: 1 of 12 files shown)


chrisli30:
Both Copilot findings addressed in b612b2f (force-pushed after rebase onto current staging which now has #545):

**Root cause** — the operator's config parser is SEPARATE from the aggregator/gateway one (own `OperatorConfig` struct in `operator/operator.go`). The `deriveWsURL` helper I added only ran on the aggregator side, so removing `eth_ws_url:` from `operator-railway.yaml` / `operator-ethereum-railway.yaml` would have left `EthWsUrl` empty at startup and the WS dial in `NewOperator` (operator/operator.go:513) would have failed.

**Fix**
- Exported `DeriveWsURL` from `core/config` (was lowercase / package-internal)
- `NewOperatorFromConfig` now derives `eth_ws_url` for all three layers the operator reads — top-level `EthWsUrl`, `TargetChain.EthWsUrl`, and every entry in `Chains[].EthWsUrl`. Empty in → derived from the matching `eth_rpc_url`; explicit value wins.
- Added `operator/derive_ws_url_test.go` with two regression tests: (a) all three layers derive when missing, (b) explicit `eth_ws_url` values are preserved.

```
go test -count=1 -run "TestDeriveWsURL|TestOperatorConfig" ./core/config/ ./operator/ — passes
```

PR #556, opened 2026-06-05
chore: derive version semver from git tag at build time

chrisli30:
## Summary

Fixes the stale-version bug surfaced by Position D's wallet-popup display. The chain of stale was:

1. `version/version.go` hardcoded `semver = "3.0.0"` as the source-level default
2. `make build` only injected `-X version.revision`, never `-X version.semver`
3. `Dockerfile` likewise only consumed `ARG GIT_REVISION`, never a semver build-arg
4. Railway's source-build path (no Makefile, no ldflags overrides) therefore reported `version: "3.0.0"` for every release from v3.0.0 through v3.2.0

Position D's SDK now stamps the user's signed message with whatever `/health` returns. So a stale version isn't only a telemetry annoyance anymore — it shows up in the wallet popup of every user who signs in. Real production gateways were telling users they were signing against `3.0.0` when they were actually talking to v3.2.0.

## The fix in three places

So the default value never has to be the right answer:

| File | Change |
|---|---|
| `Makefile` | New `SEMVER` + `REVISION` vars derived from `git describe --tags --abbrev=0` and `git rev-parse --short HEAD` (with `echo dev`/`echo unknown` fallbacks). Shared `VERSION_LDFLAGS` consumed by both `build` and `build-prod` targets — can't drift apart. Override-friendly: `make build SEMVER=4.0.0` still works. |
| `Dockerfile` | Same derivation runs inside the builder stage (`.git` is in the build context per `.dockerignore`). Optional `ARG SEMVER` / `ARG REVISION` let CI inject explicit overrides via `--build-arg`, but the no-arg path does the right thing — which is the path Railway uses today. |
| `version/version.go` | Default bumped `3.0.0` → `3.2.0` as a safety net for builds that bypass both Makefile and Dockerfile (e.g. a one-off `go build ./...`). Comment now describes the actual policy. |

## Verification

```bash
$ make -n build
go build -o ./out/ap -ldflags "-X 'github.com/AvaProtocol/EigenLayer-AVS/version.semver=3.2.0' -X 'github.com/AvaProtocol/EigenLayer-AVS/version.revision=212dc07'"

$ make build
$ go version -m ./out/ap | grep ldflags
  build  -ldflags="-X 'github.com/AvaProtocol/EigenLayer-AVS/version.semver=3.2.0' -X 'github.com/AvaProtocol/EigenLayer-AVS/version.revision=212dc07'"

$ ./out/ap aggregator --config=config/gateway-dev.yaml &
$ curl -s http://localhost:8080/api/v1/health
{"chainId":11155111,"status":"ok","version":"3.2.0"}
```

(was `"version":"3.0.0"` pre-fix.)

## Test plan

- [x] `make -n build` shows the derived ldflags
- [x] `make build` produces a binary stamped with both via `go version -m`
- [x] Local `/health` returns `3.2.0` after rebuild
- [ ] Railway redeploy of staging — verify `/health` reports the actual gateway version (whatever `git describe` says on the deploy commit), not the hardcoded `3.2.0`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,11 +13,18 @@ RUN go mod download
 # Copy source
 COPY . .
 
-# Build with version info
-ARG GIT_REVISION=unknown
-RUN go build \
-    -o ./ap \
-    -ldflags "-X github.com/AvaProtocol/EigenLayer-AVS/version.revision=${GIT_REVISION}"
+# Build with version metadata. SEMVER + REVISION are derived from git
+# inside the builder (the build context retains .git — see .dockerignore).
+# CI workflows can still inject explicit overrides via --build-arg, e.g.
+# `docker build --build-arg SEMVER=3.2.0 --build-arg REVISION=abc1234`.
+ARG SEMVER=
+ARG REVISION=
+RUN SEMVER="${SEMVER:-$(git describe --tags --abbrev=0 2>/dev/null | sed 's/^v//' || echo dev)}" && \
+    REVISION="${REVISION:-$(git rev-parse --short HEAD 2>/dev/null || echo unknown)}" && \
+    echo "Building with semver=${SEMVER} revision=${REVISION}" && \
+    go build \
+      -o ./ap \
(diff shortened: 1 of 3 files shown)


chrisli30:
@claude, address copilot’s review

PR #559, opened 2026-06-05
fix: emit canonical REST camelCase values for NodeCOGS / ValueFee enums

chrisli30:
## Summary

Fee estimator emits canonical REST camelCase values, not snake_case. Sourced from new shared constants so it stays that way.

| Field | Before | After |
|---|---|---|
| `NodeCOGS.cost_type` | `"wallet_creation"` | `"walletCreation"` |
| `ValueFee.classification_method` | `"rule_based"` | `"ruleBased"` |
| `NodeCOGS.cost_type` (gas) | `"gas"` (already correct, now constant-sourced) | `CostTypeGas` |

OpenAPI advertises these as camelCase enums (`NodeCOGSCostType ∈ {gas, externalApi, walletCreation}`, `ValueFeeClassificationMethod ∈ {ruleBased, llm}`). The REST mapper in `aggregator/rest/mapping/fees.go` translates field **names** via protojson camelCase but passes field **values** through verbatim, so the engine's raw snake_case strings were leaking straight to REST clients. Latent through PR #538; surfaced now that the [ava-sdk-js E2E suite is actually executing](https://github.com/AvaProtocol/ava-sdk-js/commit/c2b4f01) against the gateway.

## Test that caught it

`tests/v4/executions/estimateFees.test.ts` in ava-sdk-js:

```ts
expect(["gas", "externalApi", "walletCreation"]).toContain(c.costType);
// before this PR:
//   Expected value: "wallet_creation"
//   Received array: ["gas", "externalApi", "walletCreation"]
```

`classification_method` drift was latent — no SDK test exercises it yet — but it would have surfaced the same way the moment one did.

## What's in the PR

- **`core/taskengine/fee_enums.go`** (new) — canonical string constants for the four enum classes the engine emits into REST-visible proto string fields. `CostType{Gas, ExternalAPI, WalletCreation}`, `ClassificationMethod{RuleBased, LLM}`, `DiscountType{NewUser, Volume, Promotional, BetaProgram}`. The discount-type constants are preemptive — no emit sites today (`Discounts: []` in `fee_estimator.go:238`), but landing them now means future wiring spells the right value the first time.

- **`core/taskengine/fee_estimator.go`** — every raw-string emit (six sites) replaced with the constants.

- **`aggregator/rest/mapping/fee_enums_drift_test.go`** (new) — fail-build test that compares the engine constants against `generated.*` enum values in this package's `generated` sibling. Fails loudly if either side adds/renames/removes a value without updating the other. **This is the load-bearing piece** — it's what prevents the next engineer from forgetting to mirror an OpenAPI change in the constants.

- Comment cleanup so the documented canonical values match what the engine emits: `protobuf/avs.proto:1378,1389,1396`, `core/taskengine/summarizer_deterministic.go:109`, `docs/changes/20260406-fee-estimation.md` (5 spots).

## Test plan

- [x] `go test ./core/taskengine/` — passes (fee_estimator + summarizer + classifier tests)
- [x] `go test ./aggregator/rest/mapping/` — drift check + mapping tests pass
- [x] Sanity-check the drift test by temporarily reverting `CostTypeWalletCreation` to `"wallet_creation"` — the test failed with a clear diff (`+ wallet_creation / - walletCreation`), then I restored the change.
- [ ] After merge: confirm `publish-dev-docker.yml` rebuilds `avs-dev:latest` (the proto comment edit trips the `paths: ['protobuf/avs.proto']` filter — no manual `workflow_dispatch` needed)
- [ ] After image rebuild: the ava-sdk-js `E2E / executions` matrix shard goes green ([currently red on this one assertion](https://github.com/AvaProtocol/ava-sdk-js/actions/runs/27031353496))

## Downstream impact

- **REST clients on the v4 SDK contract** — including ava-sdk-js, the only in-tree consumer today — start receiving the correct `walletCreation` / `ruleBased` values. The ava-sdk-js `executions` shard goes green at that point.
- **gRPC clients pinning the snake_case `cost_type` value** would break. There is no such client in-tree: the v3 archive doesn't read fee fields, and v4 talks REST exclusively.

## Proto change

Comment-only — `cost_type`, `classification_method`, and `discount_type` are `string` fields, not protobuf enums, so no `protoc-gen` regen is needed. The proto-touch is deliberate: `publish-dev-docker.yml` auto-rebuilds `avs-dev:latest` on push to staging when `protobuf/avs.proto` changes, which is how the SDK's `:latest` pull picks up the fix.

### DIFF
--- a/aggregator/rest/mapping/fee_enums_drift_test.go
+++ b/aggregator/rest/mapping/fee_enums_drift_test.go
@@ -1,0 +1,86 @@
+package mapping
+
+import (
+	"sort"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/AvaProtocol/EigenLayer-AVS/aggregator/rest/generated"
+	"github.com/AvaProtocol/EigenLayer-AVS/core/taskengine"
+)
+
+// TestFeeEnumsDriftFromOpenAPI guards the engine-side canonical
+// enum constants (core/taskengine/fee_enums.go) against the
+// OpenAPI-generated REST enum types in this package's `generated`
+// sibling. If either side adds/removes/renames a value, this test
+// fails — which is the intended catch.
+//
+// Background: NodeCOGS.cost_type, ValueFee.classification_method,
+// and FeeDiscount.discount_type are declared as protobuf `string`
+// fields, not protobuf enums. protojson camelCases field NAMES but
+// passes field VALUES through verbatim. So the canonical wire
+// value is whatever the engine literally writes into the field —
+// and the only way to keep that aligned with the REST contract
+// (which DOES typed-enum these fields) is to source the strings
+// from one canonical place and assert the sets match.
+func TestFeeEnumsDriftFromOpenAPI(t *testing.T) {
+	cases := []struct {
+		name    string
+		openAPI []string
+		engine  []string
+	}{
+		{
+			name: "NodeCOGS.costType",
+			openAPI: []string{
+				string(generated.Gas),
+				string(generated.ExternalApi),
+				string(generated.WalletCreation),
+			},
+			engine: []string{
+				taskengine.CostTypeGas,
+				taskengine.CostTypeExternalAPI,
+				taskengine.CostTypeWalletCreation,
+			},
+		},
+		{
+			name: "ValueFee.classificationMethod",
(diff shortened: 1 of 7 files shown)


chrisli30:
@claude, review

PR #582, opened 2026-06-14
fix(trigger): rollback partial EventTrigger subscriptions and surface rebuild metrics

### LINKED ISSUE #570: Operator may silently drop event subscriptions after transient resubscribe failure
## Summary

The mainnet operator's `EventTrigger` reconnection loop can silently drop individual subscriptions after a transient upstream RPC failure during the resubscribe step, leaving the operator unable to detect events for the affected task(s) until something else triggers a full reconnect cycle.

Surfaced by Sentry issue [`EIGENLAYER-AVS-S`](https://ava-protocol.sentry.io/issues/EIGENLAYER-AVS-S) (`websocket: close 1011 (internal server error): Bad Gateway`, 18 events on the mainnet operator `0xc6B87cc9e85b07365b6aBEfff061F237F7cf7Dc3` between March and May). The visible Sentry log was downgraded as part of the cleanup, but the underlying behavior — silently dropping a subscription — is still present.

## Code location

[`core/taskengine/trigger/event.go:775-805`](https://github.com/AvaProtocol/EigenLayer-AVS/blob/main/core/taskengine/trigger/event.go#L775-L805) — the resubscribe step inside the reconnection loop:

```go
// Create new subscriptions
for i, queryInfo := range newQueries {
    timeoutCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
    sub, subErr := t.wsEthClient.SubscribeFilterLogs(timeoutCtx, queryInfo.Query, logs)
    cancel()
    if subErr != nil {
        t.logger.Error("❌ Failed to recreate subscription", "index", i, "error", subErr)
        continue  // ← subscription i is now permanently missing until next full reconnect
    }
    ...
}
```

## Scenario

1. Upstream RPC has a transient blip (502 / 1011 / 10s timeout for one filter).
2. The errorCh handler fires → `retryConnectToRpc()` re-establishes the WebSocket.
3. The loop iterates through `newQueries` to recreate each subscription.
4. One (or more) of those `SubscribeFilterLogs` calls fails — upstream still flaky.
5. We log + `continue` and move on. The remaining subscriptions get created fine.
6. From this point on, the operator has N-1 (or fewer) active subscriptions for the rest of the connection's lifetime — that task's events are simply not seen.
7. Self-heal only happens when another subscription errors out and triggers a full re-build of all subs (because the receive on `errorCh` is what drives the rebuild).

## Why it matters

- Silent failure: no metric, no alert, no aggregator-side signal. The task just stops triggering.
- Affects mainnet operator (production). Hasn't recurred in Sentry since May 16 because the upstream has been stable, but the latent bug is unchanged.
- The Sentry log on this line was just downgraded `Error → Warn` to stop the noise — so future occurrences won't even page anymore.

## Proposed approaches (pick one, or combine)

1. **Per-subscription retry with bounded attempts** — try `SubscribeFilterLogs` 2–3 times with backoff before giving up on a sub.
2. **Escalate to a full reconnect on any resubscribe failure** — if any sub fails to recreate, drop the whole WS connection and re-enter the reconnection loop from scratch. Simple and matches existing error-driven recovery.
3. **Track active vs. desired subscription count and expose a metric** — `subscriptions_active{taskID}` / `subscriptions_desired{taskID}` so missing subs are visible in dashboards even if option 1 or 2 also fails.
4. **(Minimum, no-fix option)** — at least restore the `Error` log here, so we don't lose the only signal we have today. But this would re-introduce the Sentry noise the March cleanup intentionally removed.

(2)+(3) is probably the right combination: rebuild on any failure (since the failure window is short-lived anyway), and emit a metric so we would notice if the rebuild itself starts looping.

## Acceptance criteria

- When upstream RPC transiently fails during the resubscribe step, the operator either retries that subscription or escalates to a full reconnect — never silently runs with fewer subs than desired.
- A metric (Prometheus counter or gauge) is exposed showing active vs. desired subscription counts so silent gaps are detectable on the telemetry dashboard.
- A regression test or integration test covers the case where `SubscribeFilterLogs` returns an error during resubscribe.

## Related

- Commit `ca69ad1` ("downgrade transient connection errors from Error to Warn") fixed the surrounding Sentry noise but did not touch this behavioral concern.
- Sentry issue `EIGENLAYER-AVS-S` (now resolved).

Antrikshgwal:
Fixes #570 

This PR fixes the EventTrigger resubscribe path so a single SubscribeFilterLogs failure no longer leaves the trigger running in a silent partial state. Incremental subscription updates now roll back any newly created subscriptions and escalate into the reconnect loop, which retries the full rebuild instead of continuing with missing subscriptions.

It also wires chain-level subscription telemetry into the operator so we can see active vs desired subscription counts and count rebuild failures. A regression test covers the rollback behaviour for both the incremental add path and the full rebuild path.

Validation:
- Targeted tests passed for TestRebuildSubscriptionsDoesNotCommitPartialSet and TestAddSubscriptionsRollsBackOnError
- Touched files compiled cleanly

### DIFF
--- a/core/taskengine/trigger/event.go
+++ b/core/taskengine/trigger/event.go
@@ -92,6 +92,8 @@ type QueryInfo struct {
 	MaxEventsPerBlock uint32
 }
 
+type subscribeFilterLogsFunc func(context.Context, ethereum.FilterQuery, chan<- types.Log) (ethereum.Subscription, error)
+
 type EventTrigger struct {
 	*CommonTrigger
 
@@ -136,6 +138,10 @@ type EventTrigger struct {
 
 	// Token enrichment service for enriching Transfer events
 	tokenEnrichmentService *taskengine.TokenEnrichmentService
+
+	subscribeFilterLogs         subscribeFilterLogsFunc
+	setSubscriptionCounts       func(active, desired int)
+	incSubscriptionRebuildError func()
 }
 
 type SubscriptionInfo struct {
@@ -188,6 +194,9 @@ func NewEventTrigger(o *RpcOption, triggerCh chan TriggerMetadata[EventMark], lo
 	if err != nil {
 		panic(err)
 	}
+	b.subscribeFilterLogs = func(ctx context.Context, query ethereum.FilterQuery, logs chan<- types.Log) (ethereum.Subscription, error) {
+		return b.wsEthClient.SubscribeFilterLogs(ctx, query, logs)
+	}
 
 	// Use global TokenEnrichmentService or initialize if not set
 	if taskengine.GetTokenEnrichmentService() == nil {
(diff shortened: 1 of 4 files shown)


chrisli30:
Hi @Antrikshgwal,

Thanks for the PR, let us take a look!

will-dz:
@Antrikshgwal, thanks for this. The all-or-nothing rollback is what #570 needed. Two small asks before merge:

1. `git rebase staging` and **re-target the PR to `staging`**. We land PRs on staging, not main directly (main only updates via the release flow).

2. Collapse the double retry around `retryConnectToRpc`. That helper already loops on dial failure and only returns nil, so the new if err != nil { break } branch is unreachable and we end up with two stacked backoffs. Could you simplify it to a single backoff path?

3. No need to worry about the unit test failures. We will re-run them from staging. 

Everything else looks good once no.1 and no.2 are addressed.

Antrikshgwal:
> @Antrikshgwal, thanks for this. The all-or-nothing rollback is what #570 needed. Two small asks before merge:
> 
> 1. `git rebase staging` and **re-target the PR to `staging`**. We land PRs on staging, not main directly (main only updates via the release flow).
> 2. Collapse the double retry around `retryConnectToRpc`. That helper already loops on dial failure and only returns nil, so the new if err != nil { break } branch is unreachable and we end up with two stacked backoffs. Could you simplify it to a single backoff path?
> 3. No need to worry about the unit test failures. We will re-run them from staging.
> 
> Everything else looks good once no.1 and no.2 are addressed.

Done. The reconnect loop in event.go now calls retryConnectToRpc once per iteration, skips the unreachable error branch, and keeps a single outer backoff only for rebuild failures. I also added a shutdown guard after reconnect so we don’t try to rebuild during teardown.

chrisli30:
> > @Antrikshgwal, thanks for this. The all-or-nothing rollback is what #570 needed. Two small asks before merge:
> > 
> > 1. `git rebase staging` and **re-target the PR to `staging`**. We land PRs on staging, not main directly (main only updates via the release flow).
> > 2. Collapse the double retry around `retryConnectToRpc`. That helper already loops on dial failure and only returns nil, so the new if err != nil { break } branch is unreachable and we end up with two stacked backoffs. Could you simplify it to a single backoff path?
> > 3. No need to worry about the unit test failures. We will re-run them from staging.
> > 
> > Everything else looks good once no.1 and no.2 are addressed.
> 
> Done. The reconnect loop in event.go now calls retryConnectToRpc once per iteration, skips the unreachable error branch, and keeps a single outer backoff only for rebuild failures. I also added a shutdown guard after reconnect so we don’t try to rebuild during teardown.

Thanks, look good to merge!

PR #584, opened 2026-06-14
feat: route withdraw balance reads through chain workers (Phase 4a)

chrisli30:
## What

Phase 4a of the worker-RPC delegation (`docs/changes/20260612-delegate-chain-rpc-to-workers.md`). `ExecuteWithdraw`'s preflight **balance validation** was the last live gateway path issuing direct execution-chain RPC. The send path already routed through the worker (`sendUserOpViaWorker` → `ChainWorker.ExecuteUserOp`); this moves the balance reads off the gateway too.

## Changes

- **`worker.proto`**: add `GetBalance` (native wei, wraps `ethclient.BalanceAt`) and `GetTokenBalance` (ERC-20, wraps `erc20.BalanceOf`); implemented both in `worker/server.go`.
- **`ChainStateReader`**: add `GetBalance` + `GetTokenBalance` to the interface and both implementations (`directChainStateReader`, `workerChainStateReader`).
- **`ExecuteWithdraw`**: resolves a per-chain `ChainStateReader` (worker-routed in gateway mode) for native + ERC-20 balance reads, mirroring the REST fee/nonce handlers. The direct-RPC reader is now a **fallback only** (single-chain mode / startup race).
- **Dead-code removal**: withdrawals always set `SkipReimbursement = true`, so the gas-reimbursement branch (`BuildUserOpWithPaymaster` / `EstimateGasReimbursementAmount` / `CalculateMaxWithdrawableAmount`) was unreachable. Removed it and the orphaned `CalculateMaxWithdrawableAmount` helper.
- **`resolveSmartWalletForChain`** split into a dial-free `resolveSmartWalletConfigForChain` (primary path) + the original (fallback), so the gateway no longer dials chain RPC on the live withdraw path.

## Result

No live gateway request path issues direct execution-chain RPC anymore. The per-chain `ethclient.Dial` (`smartWalletRpcByChain` + `ChainEntry.GetRPC`) remains **only** as the worker-unreachable fallback.

## Not included (Phase 4b)

Stripping the per-chain dial + the `*_RPC` env vars from the gateway Railway service is deferred to a follow-up after one prod release cycle — same migrate-then-strip safety discipline used for Phase 2→3. The top-level `eth_rpc_url` (`SEPOLIA_RPC`, the gateway's own AVS chain) stays.

## Testing

- `go build ./...`, `go vet ./aggregator/... ./worker/... ./core/taskengine/` clean.
- New `chain_state_reader_test.go` cases: `GetBalance` / `GetTokenBalance` happy-path (address propagation + base-10 round-trip) and malformed-response handling.
- Existing `TestWorkerChainStateReader*` / `TestWorkerRoutedFetcher*` pass; test fakes updated for the two new interface methods.


chrisli30:
**Update — Phase 4b scoping.** Investigated stripping the per-chain `*_RPC` env vars (the originally-planned 4b) and found it would break prod: the gateway's taskengine VM and `GetWallet` derivation dial each chain's `eth_rpc_url` **directly** (CREATE2 derivation at `engine.go:1275`, contractRead `eth_call`, block context, `FilterLogs`, simulation storage probes) — entirely separate from the `smartWalletRpcByChain` map this PR freed up. `getWallet` even hard-errors on an empty per-chain RPC URL.

So the env-var strip is **not** a config change — it requires routing the full VM read/execute surface through workers first. This PR now also adds `docs/changes/20260614-route-vm-execution-through-workers.md` with the complete direct-RPC inventory and a 7-PR phased plan (interface-extraction via a `ChainReader`, reusing the Phase 3–4a registry), ending in the env-var strip as the final phase. The old doc's Phase 4b section is marked superseded.

This PR remains correct and self-contained as Phase 4a (withdraw balance reads → workers).

chrisli30:
Addressed Copilot's review in 1d7a135:

1. **`resolveSmartWalletConfigForChain` requiring a connected worker** — fixed. Switched from `GetWorker` (errors when the worker gRPC is disconnected) to `GetChainConfig` (config-only lookup). Config resolution no longer depends on worker connectivity, so the direct-RPC fallback is actually reachable when a worker is briefly down.

2. **`chain_id=0` reintroducing gateway-side RPC** — intentionally left as-is. For `chain_id=0` the fallback reads via `r.smartWalletRpc`, which is the gateway's own **AVS-chain** client (the one connection it legitimately keeps), and that matches the top-level `smart_wallet` config `chain_id=0` resolves to. Resolving `0 → registry default` would point the reader at `chains[0]` (mainnet) while the config stays the top-level chain (Sepolia) — a chain mismatch on the balance read. So the current behavior is correct; it's not a per-chain *execution* RPC being reintroduced.

3. **Unvalidated `payload.Token`** — fixed. Added `common.IsHexAddress` validation before `HexToAddress` in the ERC-20 branch (the path is reachable via gRPC, not only REST).

PR #585, opened 2026-06-14
feat: route GetWallet derivation through chain workers (VM Phase 1)

chrisli30:
## What

PR 1 of [`docs/changes/20260614-route-vm-execution-through-workers.md`](docs/changes/20260614-route-vm-execution-through-workers.md) — routing the gateway's VM/wallet-derivation layer through chain workers so the gateway eventually talks to only its own AVS chain.

`GetWallet`'s CREATE2 address derivation ([engine.go:1275](core/taskengine/engine.go)) was the **single most-hit** gateway per-chain dial — every `getWallet` call did `ethclient.Dial(swCfg.EthRpcUrl)` + `factory.getAddress()`. This routes it through the chain worker instead. Highest-traffic, lowest-risk first step.

## Worker RPC redefinition

The worker's `GetSmartWalletAddress` RPC had **zero client-side callers** (implemented but never wired up), so it's redefined to match what `GetWallet` actually needs — no back-compat constraint:

- `salt` is now a base-10 **string** (was `int64`, which can't carry a 256-bit CREATE2 salt)
- added a `factory_address` override so worker-derived addresses match the gateway's per-request / per-chain factory exactly (GetWallet supports a per-call factory override)

## Changes

- **`worker.proto` / `worker/server.go`**: `WorkerGetSmartWalletAddressReq { owner, salt(string), factory_address }`; handler parses + validates owner/salt/factory and honors the override.
- **`ChainStateReader`**: add `GetSmartWalletAddress` to the interface + both impls (direct wraps `aa.GetSenderAddressForFactory`; worker routes via gRPC).
- **`engine.go` `GetWallet`**: derive via the per-chain `ChainStateReader` when one is registered — worker-routed in gateway mode, a direct reader (no per-call dial) in single-chain mode. Falls back to the direct `ethclient.Dial` only when no reader exists. The `rpcConn==nil` test/CI mock branch is unchanged.

## Testing

- New `chain_state_reader_test.go` cases: `GetSmartWalletAddress` happy-path (owner/factory hex + base-10 salt propagation, address round-trip), nil-salt → `"0"`, malformed-address error.
- Existing `GetWallet` + `ChainStateReader` + `WorkerRoutedFetcher` tests pass.

## Next

PR 2: `CallContract` worker RPC + contractRead migration (the largest remaining surface).

### DIFF
--- a/core/taskengine/chain_state_reader.go
+++ b/core/taskengine/chain_state_reader.go
@@ -12,6 +12,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
 
+	"github.com/AvaProtocol/EigenLayer-AVS/core/chainio/aa"
 	"github.com/AvaProtocol/EigenLayer-AVS/pkg/erc20"
 	avsproto "github.com/AvaProtocol/EigenLayer-AVS/protobuf"
 )
@@ -60,6 +61,11 @@ type ChainStateReader interface {
 	// GetTokenBalance reads the raw ERC-20 balance (no decimals applied)
 	// of owner for the given token contract. Mirrors erc20.BalanceOf.
 	GetTokenBalance(ctx context.Context, token, owner common.Address) (*big.Int, error)
+
+	// GetSmartWalletAddress derives the CREATE2 smart-wallet address for
+	// (owner, salt) under the given factory on this chain. Mirrors
+	// aa.GetSenderAddressForFactory.
+	GetSmartWalletAddress(ctx context.Context, owner, factory common.Address, salt *big.Int) (common.Address, error)
 }
 
 // ---------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


chrisli30:
Addressed all four Copilot comments in b0dbd19:

1. **Empty/out-of-range salt (`worker/server.go`)** — fixed. Empty salt now defaults to `0` (proto3 omits empty strings; matches the gateway's nil→"0" convention); negative and >256-bit salts are rejected before ABI encoding.
2. **Nil derived address (`worker/server.go`)** — fixed. Guard `addr == nil` before `.Hex()`, mirroring the direct reader.
3. **Nil gRPC response (`workerChainStateReader.GetSmartWalletAddress`)** — fixed. Guard `resp == nil` before reading `resp.Address`.
4. **`%s` on `*big.Int` + nil-salt normalization (`directChainStateReader`)** — fixed. Normalize nil salt to `0` so direct and worker paths derive identical addresses; use `.String()` in the error.

Build + `go vet` + targeted tests green.

PR #586, opened 2026-06-14
feat: route contractRead eth_call through chain workers (VM Phase 2)

chrisli30:
## What

PR 2 of [`docs/changes/20260614-route-vm-execution-through-workers.md`](docs/changes/20260614-route-vm-execution-through-workers.md). The contractRead node dialed `swCfg.EthRpcUrl` directly and issued `eth_call` against execution chains from the gateway — the largest remaining direct-RPC surface. This routes it through the chain worker.

## Changes

- **`worker.proto` / `worker/server.go`**: new `CallContract` RPC (`from`/`to`/`value`/`data`/`block_number`) wrapping `ethclient.CallContract`.
- **`ChainStateReader`**: add `CallContract` to the interface + both impls (direct wraps `ethclient.CallContract`; worker-routed rejects a nil `To` and guards a nil response).
- **`ContractReadProcessor`**: field changed from `*ethclient.Client` to `ChainStateReader`. It used `CallContract`, `ChainID`, and `CodeAt` — all now on the interface. `CodeAt` sites drop the trailing block arg (interface signature); `ChainID` now returns `int64` (used only for logging / error strings).
- **`vm.go` `runContractRead`**: resolve the per-chain `ChainStateReader` (worker-routed in gateway mode), falling back to a direct dial only when no reader is registered for the chain.

## Deferred (not this PR)

`callContractMethod` ([run_node_immediately.go](core/taskengine/run_node_immediately.go)) reads `decimals()` via the package-level `rpcConn` — the gateway's **AVS-chain** client, with no chainID parameter. It's chain-ambiguous, so it moves with the event-enrichment work (PR 3/5) where chain context is threaded through. Migrating it here would require reading a token's decimals on the wrong chain.

## Testing

- New `CallContract` reader tests: To/Data propagation, From/Value/block default-empty + set-serialization, nil-To rejection.
- Existing `ContractRead*` processor tests pass (updated to wrap the test client in a `DirectChainStateReader`).

## Next

PR 3 — block context (`GetBlockNumber` / `GetBlockHeader`).

### DIFF
--- a/core/taskengine/chain_state_reader.go
+++ b/core/taskengine/chain_state_reader.go
@@ -66,6 +66,10 @@ type ChainStateReader interface {
 	// (owner, salt) under the given factory on this chain. Mirrors
 	// aa.GetSenderAddressForFactory.
 	GetSmartWalletAddress(ctx context.Context, owner, factory common.Address, salt *big.Int) (common.Address, error)
+
+	// CallContract executes a read-only contract call (eth_call) at the
+	// given block (nil = latest). Mirrors ethclient.CallContract.
+	CallContract(ctx context.Context, msg ethereum.CallMsg, blockNumber *big.Int) ([]byte, error)
 }
 
 // ---------------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


chrisli30:
Addressed both Copilot comments:

1. **`CallContract` value/block_number validation (`worker/server.go`)** — fixed. `value` now rejects negative and >256-bit inputs (it's a uint256); `block_number` rejects negative. Both fail fast with a clear error rather than deferring to a downstream JSON-RPC error.
2. **Missing `Data` assertion (`chain_state_reader_test.go`)** — fixed. Added the calldata-propagation assertion the test comment already claimed.

Build + vet + targeted tests green.

PR #590, opened 2026-06-15
feat: route immediate-trigger block read through chain workers (rpcConn A)

chrisli30:
## What

First PR of [`docs/changes/20260614-rpcconn-chain-threading.md`](docs/changes/20260614-rpcconn-chain-threading.md) — retiring the package-level `rpcConn` (the gateway's AVS-chain client) in favor of per-chain workers.

`instructOperatorImmediateTrigger` read the current block via `rpcConn` to stamp the immediate trigger for the operator. This routes it through the **task's** per-chain worker.

## Changes

- **`worker.proto` / `worker/server.go`**: new `GetBlockNumber` RPC (wraps `ethclient.BlockNumber`).
- **`ChainStateReader`**: add `GetBlockNumber` (direct + worker-routed, with nil-response guard + chain-id-wrapped error).
- **`engine.go` `instructOperatorImmediateTrigger`**: load the task first, resolve the `ChainStateReader` for `task.ChainId`, and read the current block through it. A missing reader is now an explicit error (replaces the `rpcConn == nil` guard). The block targets the **task's chain**, not a single gateway RPC.

## Scope notes

- **`runBlockTriggerImmediately`** (the other block-trigger reader) also uses `rpcConn`, but its chain source is the simulate/runTrigger trigger config — non-trivial to thread cleanly, and the existing run_node path uses the global token-service chain (which differs from `rpcConn`'s AVS chain). It moves to the next PR (rpcConn B) alongside the event-enrichment reads, where the run_node chain context is threaded properly — rather than guessing the chain here and risking a behavior change in which chain a simulate returns.
- **`utils.GetBlock`** is dead code (no callers); it'll be removed in the `rpcConn`-retirement PR.

## Testing

- New `GetBlockNumber` reader tests: happy path + chain-id-wrapped transport error.

### DIFF
--- a/core/taskengine/chain_state_reader.go
+++ b/core/taskengine/chain_state_reader.go
@@ -76,6 +76,10 @@ type ChainStateReader interface {
 	// the gateway reads — a full types.Header can't be carried over gRPC
 	// faithfully (its hash depends on every field).
 	HeaderByNumber(ctx context.Context, number *big.Int) (*BlockHeader, error)
+
+	// GetBlockNumber returns the latest block number. Mirrors
+	// ethclient.BlockNumber.
+	GetBlockNumber(ctx context.Context) (uint64, error)
 }
 
 // BlockHeader is the subset of block-header fields the gateway reads:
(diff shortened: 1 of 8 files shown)


chrisli30:
Addressed Copilot's comment: `task.ChainId == 0` (single-chain mode / legacy tasks) now falls back to the engine's default chain (`n.tokenEnrichmentService.GetChainID()`) before resolving the reader — the same key the registry uses in single-chain mode — so the immediate-trigger path no longer regresses where the old global `rpcConn` worked. The error/log use the resolved chain ID. Build + vet green.

PR #591, opened 2026-06-15
feat: thread explicit chain through block-trigger reads (rpcConn B)

chrisli30:
## What

Second PR of [`docs/changes/20260614-rpcconn-chain-threading.md`](docs/changes/20260614-rpcconn-chain-threading.md). Routes the block-trigger reads off the package-level `rpcConn` (the gateway's AVS chain) onto the **trigger's own** per-chain worker — with the chain sourced **explicitly, no engine-default fallback**.

## Changes

- **`node_types.go`**: `TaskTriggerToConfig` now emits `chain_id` for block triggers (from `BlockTrigger_Config.GetChainId`), so the immediate block read can resolve the trigger's chain.
- **`run_node_immediately.go`**: `runBlockTriggerImmediately` **requires** `chain_id` from the trigger config (new `requireChainIDFromConfig` helper — absent / zero / wrong-typed is a **hard error**, no default), resolves the per-chain `ChainStateReader`, and routes `GetBlockNumber` + `HeaderByNumber` through it.
- **`engine.go`**: `instructOperatorImmediateTrigger` **drops the `tokenEnrichmentService` default-chain fallback added in #590** — a zero `task.ChainId` is now a hard error. There is no engine-default chain; chain-scoped reads target an explicit chain or fail.
- **`worker.proto` / `chain_state_reader.go`**: extend `GetBlockHeader` + `BlockHeader` with `parent_hash`/`difficulty`/`gas_limit`/`gas_used` (the fields the block-trigger output exposes).

## Design note

Per maintainer direction: **there is no engine-default chain.** Every chain-scoped read must derive its chain explicitly from the task/trigger; a missing or zero `chain_id` hard-fails rather than silently targeting a default (which could be the wrong chain in multi-chain mode). This PR also corrects the #590 fallback to that rule.

## Testing

- New: `requireChainIDFromConfig` hard-fail matrix; extended `HeaderByNumber` field mapping.
- Full short-mode taskengine suite passes. (One failing test, `TestSimulateTask_StopLossWorkflow_Sepolia`, hard-fails at its first line on a missing `OWNER_EOA` env var — an env-gated Sepolia integration test, unrelated to this change.)

## Next

Event-enrichment reads (`parseEventWithParsedABI` timestamp, `callContractMethod` decimals) — thread the event trigger's chain (its `chain_id` emission is deferred to that PR).

### DIFF
--- a/core/taskengine/chain_state_reader.go
+++ b/core/taskengine/chain_state_reader.go
@@ -436,6 +436,14 @@ func (w *workerChainStateReader) HeaderByNumber(ctx context.Context, number *big
 	if resp == nil {
 		return nil, fmt.Errorf("worker returned nil response for GetBlockHeader on chain %d", w.chainID)
 	}
+	// A blank hash means the worker didn't populate it (version skew) —
+	// HexToHash would silently coerce "" to the zero hash and propagate an
+	// incorrect block-trigger output. Hard-fail instead. (ParentHash is
+	// genuinely zero only for genesis, but the worker always serializes it
+	// via .Hex(), so an empty string is still skew.)
+	if resp.Hash == "" || resp.ParentHash == "" {
+		return nil, fmt.Errorf("worker returned incomplete block header (hash=%q parentHash=%q) on chain %d", resp.Hash, resp.ParentHash, w.chainID)
+	}
 	difficulty, ok := new(big.Int).SetString(resp.Difficulty, 10)
 	if !ok {
 		// Difficulty is informational (post-merge it's 0); tolerate a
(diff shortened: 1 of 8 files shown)


chrisli30:
Addressed all four Copilot comments:

1 & 2. **"from RPC" error strings** — updated to reference the chain (`for chain N` / `on chain N`) since these reads are worker-routed now.
3. **`requireChainIDFromConfig` validation** — no more blind casts: non-integer `float64` is rejected, out-of-`int64`-range `uint64`/`float64` is rejected, and the result must be **positive** (negative chain IDs would become huge when cast to `uint64`).
4. **Incomplete block header** — `HeaderByNumber` now hard-fails when the worker returns an empty `hash`/`parentHash` (version skew) rather than coercing to the zero hash and emitting an incorrect block-trigger output.

New tests cover the empty-hash rejection; existing `_Latest` fixture updated to a complete header. Build + vet + targeted tests green.

PR #592, opened 2026-06-15
feat: thread chain through event-enrichment reads (rpcConn C)

chrisli30:
## What

Third PR of [`docs/changes/20260614-rpcconn-chain-threading.md`](docs/changes/20260614-rpcconn-chain-threading.md). Routes the best-effort **event-enrichment** reads off the package-level `rpcConn` (the gateway's AVS chain) onto the trigger's own per-chain worker.

Two reads migrated:
- **token decimals** — `callContractMethod` (`CallContract`)
- **event block timestamp** — `parseEventWithParsedABI` (`HeaderByNumber`)

## Skip, never default (per maintainer direction)

These are **optional, non-fatal** adornments. The new `chainReaderForEnrichment(chainID)` helper returns `nil` for `chainID <= 0` or an unregistered chain, so enrichment is **skipped** (keeps the default timestamp/decimals) rather than reading a default/wrong chain. This honors *no engine-default chain* while not hard-failing the event decode (which the block-trigger path does, because there the read *is* the output).

## Threading

- `node_types.go`: `TaskTriggerToConfig` emits `chain_id` for event triggers.
- chainID threaded through `callContractMethod`, `parseEventWithParsedABI`, `parseEventWithABI`, `runEventTriggerWithHistoricalSearch` (enrichment path only), `runEventTriggerImmediately` (soft-resolves chain from the trigger config), and `RunNodeImmediatelyRPC` (sources chain from req / node config).
- `OutputHandler.ConvertToProtobuf` interface + all 10 implementations now take `chainID` (only `ContractWriteOutputHandler` uses it, for event-log decode) — the full interface threading you approved.

## Deferred

`runEventTriggerWithHistoricalSearch`'s own `BlockNumber` + `FilterLogs` reads stay on `rpcConn` — that's the **Tenderly-first** historical-search work (PR C in the plan), gated on the simulate-fidelity decision.

## Testing

- `go build ./...`, `go vet` clean (the 9 handlers that ignore `chainID` carry an expected unused-param info).
- New `chainReaderForEnrichment` skip matrix; existing ABI-decode tests pass with `chainID 0` (enrichment skipped). Full short-mode taskengine suite green.

### DIFF
--- a/core/taskengine/abi_typing_test.go
+++ b/core/taskengine/abi_typing_test.go
@@ -65,7 +65,7 @@ func TestABIFieldTyping(t *testing.T) {
 	}
 
 	// Test the parseEventWithABI function
-	parsedData, err := engine.parseEventWithABI(mockLog, testABI, nil)
+	parsedData, err := engine.parseEventWithABI(0, mockLog, testABI, nil)
 	if err != nil {
 		t.Fatalf("Failed to parse event with ABI: %v", err)
 	}
@@ -236,7 +236,7 @@ func TestDecimalFormattingWithProperTypes(t *testing.T) {
 	// In a real scenario, this would be retrieved via RPC call
 
 	// Test without decimal formatting first
-	parsedData, err := engine.parseEventWithABI(mockLog, testABI, nil)
+	parsedData, err := engine.parseEventWithABI(0, mockLog, testABI, nil)
 	if err != nil {
 		t.Fatalf("Failed to parse event with ABI: %v", err)
 	}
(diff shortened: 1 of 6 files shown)


chrisli30:
Addressed Copilot's comment: the decimals-enrichment `else` branch now distinguishes the skip case from a real failure. When `chainReaderForEnrichment(chainID) == nil` (chain unspecified / unregistered) it logs at **Debug** — best-effort enrichment skipped, no warn-level noise per event; a genuine RPC failure (reader exists but the call failed) still **Warn**s. Build + gofmt + tests green.

PR #600, opened 2026-06-16
feat(operator): multi-chain operators per network tier + liveness hardening

chrisli30:
## Summary

Finishes the multi-chain operator migration and fixes two liveness gaps it exposes. Production ran two operators each advertising **a single chain** (`operator-ethereum`→`[1]`, `operator`→`[11155111]`), so **Base (8453), BNB (56), and Base-Sepolia (84532) tasks were orphaned** — no operator advertised their chain, so the gateway never routed or fired them.

The operator already had multi-chain machinery (`OperatorConfig.Chains`, `EffectiveChains`, per-chain trigger sets, the #581 liveness watermark), but the Railway operator configs were never moved off the legacy single-chain `eth_rpc_url: ${CHAIN_ENDPOINT}` / `target_chain` form. This wires it up.

## Intended topology
- **operator-ethereum** (mainnet tier) → monitors **all mainnets**: ethereum (1), base (8453), bnb (56).
- **operator-sepolia** (testnet tier, renamed from `operator`) → monitors **all testnets**: sepolia (11155111), base-sepolia (84532).

Each operator still *registers* against one AVS chain (`eth_rpc_url`); that's separate from the destination `chains:` it monitors.

## Part A — config
- `config/operator-ethereum-railway.yaml`: `target_chain` → `chains:` array (ethereum/base/bnb → `ETHEREUM_RPC`/`BASE_RPC`/`BNB_RPC`).
- `config/operator-railway.yaml` → **renamed** `config/operator-sepolia-railway.yaml`; `chains:` array (sepolia/base-sepolia → `SEPOLIA_RPC`/`BASE_SEPOLIA_RPC`).
- Updated Dockerfile/README references. **Railway env vars + service rename + start commands were applied out-of-band by the operator.**

## Fix 1 — a bad per-chain WS no longer crashes the whole operator
`NewBlockTrigger`/`NewEventTrigger` dial the chain's WS in their constructors and **panic** on failure. With one operator monitoring several chains, a single unreachable WS would take down *every* chain. The per-chain setup now constructs under panic-recovery and **soft-skips** the bad chain (added to `Skipped`, dropped from advertised capabilities) — same as a failed HTTP dial. WS URLs are still derived from each chain's HTTPS RPC; an explicit per-chain `eth_ws_url` can override.

## Fix 2 — an idle chain no longer un-advertises itself
The block subscription stops when a chain has no block tasks. With it stopped, no heads arrive → watermark goes stale → `supportedChainIDs()` dropped the chain, even though it was healthy and could serve event tasks. This **deadlocks coverage**: the gateway only routes tasks (incl. the first block task) to advertised chains, so a dropped idle chain never gets tasks, never restarts its subscription, never re-advertises.

Now a stale watermark counts as a stall **only when the chain has block work** (`BlockTrigger.HasBlockTasks()`); idle chains stay advertised. Adding a block task resets the watermark so a freshly (re)started subscription gets its full staleness window. The block-work probe is injected onto `ChainTriggerSet` so the liveness logic is unit-testable without a real (WS-dialing) trigger — see `TestSupportedChainIDs_IdleChainStaysAdvertised`.

## Testing
- `go build ./...`, `go vet`, and the operator + trigger suites pass.
- New/updated liveness tests cover: stalled-with-block-work → dropped; stalled-but-idle → advertised.

## Follow-up / risk
- WS derivation is relied on for Base/BNB; a provider needing a different WS host now surfaces as a *soft-skipped* chain (visible in `Skipped` / `SetChainAdvertised`) rather than a crash — set an explicit `eth_ws_url` for it.
- The watermark still only tracks block heads; an event-only chain with a dead event WS isn't caught (pre-existing, out of scope).

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -24,8 +24,8 @@ config/
 ├── worker-<chain>-dev.example.yaml      — per-chain local-dev worker template
 ├── worker-<chain>-dev.yaml              — per-chain local-dev worker, real (gitignored)
 │
-├── operator-railway.yaml                — Railway operator template
-├── operator-<chain>-railway.yaml        — per-chain operator overrides
+├── operator-sepolia-railway.yaml        — Railway operator: all TESTNET chains
+├── operator-ethereum-railway.yaml       — Railway operator: all MAINNET chains
 └── operator-<chain>.yaml                — local operator configs (gitignored symlinks)
 ```
 
@@ -35,7 +35,8 @@ config/
 |---|---|
 | Production aggregator on Railway | `gateway-railway.yaml` |
 | Production worker for chain N on Railway | `worker-<chain>-railway.yaml` |
-| Production operator on Railway | `operator-railway.yaml` (+ optional per-chain override) |
+| Production testnet operator on Railway | `operator-sepolia-railway.yaml` (monitors all testnet chains) |
+| Production mainnet operator on Railway | `operator-ethereum-railway.yaml` (monitors all mainnet chains) |
 | Local dev gateway | `gateway-dev.yaml` (copy from `gateway-dev.example.yaml`, fill in secrets) |
 | Local dev worker for chain N | `worker-<chain>-dev.yaml` (same copy pattern) |
 | Operating drill / rehearsal | `gateway-dev-rehearsal.yaml` |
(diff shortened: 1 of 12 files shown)


chrisli30:
@claude, review

PR #602, opened 2026-06-16
feat: multi-chain operators per network tier + liveness hardening

chrisli30:
## Release: multi-chain operators + EventCondition fix

Promotes the staged work to production so the **multi-chain operator** changes can be tested after deploy.

### Operator: multi-chain per network tier (#600)
Each operator now monitors an **array** of chains instead of a single chain, fixing orphaned Base/BNB/Base-Sepolia tasks:
- **operator-ethereum** → all mainnets (ethereum 1, base 8453, bnb 56)
- **operator-sepolia** (renamed from `operator`) → all testnets (sepolia 11155111, base-sepolia 84532)

Plus two liveness-hardening fixes the multi-chain setup exposed:
- A bad per-chain WS endpoint now **soft-skips that chain** instead of crashing the whole operator (panic-recovery around trigger construction).
- An **idle chain** (no block tasks) no longer un-advertises itself — `supportedChainIDs()` treats a stale head watermark as a stall only when the chain has block work, breaking the coverage deadlock where a dropped idle chain could never be routed tasks.
- Review follow-ups: primary-chain RPC globals (the `chains:`-only config left `TargetChain` empty), trigger-client leak on the soft-skip path, and shared channel buffers scaled by chain count.

### Fix: EventCondition.value is a string (#601)
Corrects `EventCondition.value` handling — it's a string, not an object.

### Pre-merge checks
- `make storage-check` vs `origin/main`: ✅ no breaking storage changes (key templates + model structs unchanged).
- No version.go conflict — main's post-release bump (3.8.0→3.9.0) doesn't overlap with staged changes.

### Post-deploy
Operator changes (multi-chain advertisement) will be validated after the release deploys: `operator-ethereum` should advertise `[1, 8453, 56]` and `operator-sepolia` `[11155111, 84532]`, clearing the Base/BNB/Base-Sepolia orphans. Railway env vars + service rename already applied.

### DIFF
--- a/aggregator/rest/generated/types.gen.go
+++ b/aggregator/rest/generated/types.gen.go
@@ -550,7 +550,13 @@ type EventCondition struct {
 	FieldName string                 `json:"fieldName"`
 	FieldType *string                `json:"fieldType,omitempty"`
 	Operator  EventConditionOperator `json:"operator"`
-	Value     map[string]interface{} `json:"value"`
+
+	// Value Value to compare against, encoded as a string. The operator
+	// parses it according to `fieldType` (e.g. `int256` / `uint256`
+	// → big.Int, `address` → checksummed hex, `bool` →
+	// "true"/"false"). Matches the proto `EventCondition.value`,
+	// which is also a string.
+	Value string `json:"value"`
 }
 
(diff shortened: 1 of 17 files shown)


chrisli30:
@claude, review

PR #603, opened 2026-06-16
feat: thread request context to gateway worker-routed chain calls

### LINKED ISSUE #599: Propagate request context through gateway worker-routed chain calls
## Summary

Several gateway → worker chain-state calls pass `context.Background()` instead of propagating the incoming request context. As a result, request cancellation (client disconnect, HTTP/gRPC timeout) cannot interrupt them. Each call is still bounded by the worker reader's own `withTimeout`, so this is a robustness/latency improvement, not a correctness bug.

Surfaced by the `@claude` review of #598; classed as a follow-up there.

## Why it wasn't done in #598

None of the enclosing functions currently take a `ctx` parameter, so threading one is a cross-cutting signature refactor up to the gRPC handlers. The benefit is marginal because `workerChainStateReader` already applies `withTimeout` to every call, bounding worst-case duration regardless. It was deliberately deferred to keep the #598 release PR focused.

## Call sites (worker-routed calls using `context.Background()`)

| Site | Enclosing function | Worker call |
|---|---|---|
| `core/taskengine/executor.go:899` | `validateWalletOwnership` | `GetSmartWalletAddress` |
| `core/taskengine/executor.go:958` | `validateDerivedWallet` | `FindMatchingWalletSalt` |
| `core/taskengine/engine.go:1273` | `GetWallet` | `GetSmartWalletAddress` |
| `core/taskengine/engine.go:3102` | `instructOperatorImmediateTrigger` | `GetBlockNumber` |
| `core/taskengine/run_node_immediately.go:117` | `runBlockTriggerImmediately` | `GetBlockNumber` |
| `core/taskengine/run_node_immediately.go:137` | `runBlockTriggerImmediately` | `HeaderByNumber` |
| `core/taskengine/run_node_immediately.go:2575` | `runProcessingNodeWithInputs` | `FindMatchingWalletSalt` |

(Line numbers are approximate and may drift; the functions are the stable anchors.)

## Proposed approach

Thread the incoming request `ctx` from the gRPC handler layer down through these functions, replacing `context.Background()` at the worker-call sites. Each enclosing function gains a leading `ctx context.Context` parameter; update callers up to the handler that already holds the request context.

Keep the worker reader's `withTimeout` as the lower bound (so a missing/`Background` ctx still can't hang); request cancellation then becomes an additional, earlier cancellation signal.

## Acceptance criteria

- [ ] The seven sites above propagate a request-scoped `ctx` instead of `context.Background()`.
- [ ] A cancelled request context interrupts an in-flight worker call (covered by a test).
- [ ] No behavioral change when the context is not cancelled (worker `withTimeout` still applies).

## Notes

- Scope is the gateway's worker-routed path; operator-side triggers are out of scope.
- Pairs naturally with Phase 4 (deleting the dead direct-dial fallbacks), since both touch the same functions.

Antrikshgwal:
Fixes #599 

## Summary

Seven gateway → chain-worker calls passed `context.Background()`, so request cancellation (client disconnect, HTTP timeout) could not interrupt them. Each call was bounded only by the worker reader's own `withTimeout`.

This change threads the incoming request context down to all seven sites, making cancellation an additional, earlier signal. `withTimeout` remains the lower bound, so behaviour is unchanged when the context is never cancelled.

## Sites Updated

| Enclosing Function | Worker Call |
| --- | --- |
| `executor.go` → `validateWalletOwnership` | `GetSmartWalletAddress` |
| `executor.go` → `validateDerivedWallet` | `FindMatchingWalletSalt` |
| `engine.go` → `GetWallet` | `GetSmartWalletAddress` |
| `engine.go` → `instructOperatorImmediateTrigger` | `GetBlockNumber` |
| `run_node_immediately.go` → `runBlockTriggerImmediately` | `GetBlockNumber` |
| `run_node_immediately.go` → `runBlockTriggerImmediately` | `HeaderByNumber` |
| `run_node_immediately.go` → `runProcessingNodeWithInputs` | `FindMatchingWalletSalt` |

## Why the Implementation Differs from the Issue's Proposed Approach

The issue proposed:

> "Each enclosing function gains a leading `ctx context.Context` parameter; update callers up to the handler that already holds the request context."

Taken literally, this would require changing the signatures of several public methods:

- `GetWallet`
- `RunNodeImmediately`
- `RunNodeImmediatelyRPC`
- `RunTriggerRPC`
- `RunTask`
- `TriggerWorkflow`

Those methods have roughly 150 call sites, almost all in tests. The result would be a large, mostly mechanical diff and a breaking change to exported APIs for what the issue itself classifies as a robustness and latency improvement rather than a correctness fix.

Instead, this implementation takes a narrower approach:

- **Private internals accept `ctx` directly** (`validate*`, `runTriggerImmediately`, `runBlockTriggerImmediately`, `runProcessingNodeWithInputs`, `instructOperatorImmediateTrigger`, and the `GetWallet` implementation). Their callers are internal, so no API churn is required.
- **Public methods retain their existing signatures** as thin wrappers that delegate using `context.Background()`. The actual implementation is moved to new context-aware siblings:
  - `GetWalletWithContext`
  - `RunTaskWithContext`
  - `RunNodeImmediatelyWithContext`
  - `RunNodeImmediatelyRPCWithContext`
  - `RunTriggerRPCWithContext`
  - `TriggerWorkflowWithContext`
- **The four REST handlers** (wallets, triggers, nodes, and workflows)—the actual request-context entry points—call the `...WithContext` variants using `ctx.Request().Context()`.

This achieves the same result at the seven worker-call sites while:

- Preserving backward compatibility for exported APIs.
- Avoiding widespread test churn (only three internal callers required updates).
- Accurately modeling the async queue path (`Perform`), which has no request context and therefore continues to use `context.Background()`.

## Testing

`TestRunBlockTriggerImmediately_ContextCancellationInterruptsWorkerCall` registers a fake `ChainStateReader` whose `GetBlockNumber` method signals startup and then blocks on `ctx.Done()`.

The test:

1. Starts the worker call.
2. Cancels the request context while the call is in flight.
3. Verifies that the call returns an error wrapping `context.Canceled`.

Under the previous `context.Background()` implementation, the call would have remained blocked until the assertion timeout, demonstrating that cancellation now propagates correctly.

Additionally:

- `go build ./...` passes.
- `go vet` passes on all touched packages.

## Acceptance Criteria

- [x] All seven sites propagate a request-scoped `ctx` instead of `context.Background()`.
- [x] A canceled request context interrupts an in-flight worker call (test added).
- [x] No behavior changes when the context is not canceled (`withTimeout` still applies).

### DIFF
--- a/core/taskengine/engine.go
+++ b/core/taskengine/engine.go
@@ -3419,6 +3419,15 @@ func (n *Engine) TriggerWorkflowWithContext(ctx context.Context, user *model.Use
 // default. Per-node / per-trigger chain_id (set on the node configs) takes
 // precedence over this value.
 func (n *Engine) SimulateWorkflow(user *model.User, trigger *avsproto.TaskTrigger, nodes []*avsproto.TaskNode, edges []*avsproto.TaskEdge, inputVariables map[string]interface{}, chainIDs ...int64) (*avsproto.Execution, error) {
+	return n.SimulateWorkflowWithContext(context.Background(), user, trigger, nodes, edges, inputVariables, chainIDs...)
+}
+
+// SimulateWorkflowWithContext is SimulateWorkflow with a caller-supplied
+// context. The REST simulate handler passes the request context so a client
+// disconnect / timeout cancels the worker-routed trigger reads (e.g. a
+// BlockTrigger's GetBlockNumber / HeaderByNumber) reached via
+// runTriggerImmediately.
+func (n *Engine) SimulateWorkflowWithContext(ctx context.Context, user *model.User, trigger *avsproto.TaskTrigger, nodes []*avsproto.TaskNode, edges []*avsproto.TaskEdge, inputVariables map[string]interface{}, chainIDs ...int64) (*avsproto.Execution, error) {
 	var chainID int64
 	if len(chainIDs) > 0 {
 		chainID = chainIDs[0]
@@ -3533,7 +3542,7 @@ func (n *Engine) SimulateWorkflow(user *model.User, trigger *avsproto.TaskTrigge
 	// Step 1: Start timing BEFORE trigger execution (consistent with node timing)
 	triggerStartTime := time.Now()
 
-	triggerOutput, err := n.runTriggerImmediately(context.Background(), triggerTypeStr, triggerConfig, inputVariables)
+	triggerOutput, err := n.runTriggerImmediately(ctx, triggerTypeStr, triggerConfig, inputVariables)
 	if err != nil {
 		return nil, fmt.Errorf("failed to simulate trigger: %w", err)
 	}
--- a/core/taskengine/engine.go
+++ b/core/taskengine/engine.go
@@ -1194,6 +1194,14 @@ func (n *Engine) resolveUserChainID(user *model.User) int64 {
 // resolveUserChainID — REST callers route by their JWT `aud` chain,
 // gRPC callers fall back to the gateway's default chain.
 func (n *Engine) GetWallet(user *model.User, payload *avsproto.GetWalletReq) (*avsproto.GetWalletResp, error) {
+	return n.GetWalletWithContext(context.Background(), user, payload)
+}
+
+// GetWalletWithContext is GetWallet with a caller-supplied context. REST
+// handlers pass the request context so a client disconnect / timeout cancels
+// the worker-routed CREATE2 derivation; the worker reader's withTimeout still
+// bounds the call when ctx carries no deadline.
+func (n *Engine) GetWalletWithContext(ctx context.Context, user *model.User, payload *avsproto.GetWalletReq) (*avsproto.GetWalletResp, error) {
 	chainID := n.resolveUserChainID(user)
 	// Resolve the per-chain smart wallet config. In single-chain mode
 	// this returns n.smartWalletConfig (the gateway default); in
@@ -1270,7 +1278,7 @@ func (n *Engine) GetWallet(user *model.User, payload *avsproto.GetWalletReq) (*a
 		// gateway mode this is worker-routed, so the gateway issues no
 		// direct chain RPC for wallet derivation; in single-chain mode it's
 		// a direct reader that reuses the existing client (no per-call dial).
-		addr, deriveErr := reader.GetSmartWalletAddress(context.Background(), user.Address, factoryAddr, saltBig)
+		addr, deriveErr := reader.GetSmartWalletAddress(ctx, user.Address, factoryAddr, saltBig)
 		if deriveErr != nil {
 			n.logger.Error("GetWallet: factory.getAddress() failed (worker-routed)",
 				"chain_id", chainID, "owner", user.Address.Hex(), "factory", factoryAddr.Hex(),
@@ -3080,7 +3088,7 @@ func (n *Engine) GetWorkflow(user *model.User, taskID string) (*model.Workflow, 
 
 // instructOperatorImmediateTrigger instructs the connected operator to trigger the task immediately
 // with the current block number, bypassing the normal interval waiting
-func (n *Engine) instructOperatorImmediateTrigger(taskID string) error {
+func (n *Engine) instructOperatorImmediateTrigger(ctx context.Context, taskID string) error {
 	// Load the task first so the current-block read targets the task's
 	// chain (via its per-chain worker), not a single gateway RPC.
 	task, err := n.GetWorkflowByID(taskID)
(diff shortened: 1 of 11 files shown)


chrisli30:
Thanks @Antrikshgwal, let us take a look at this resolution. 👁️

core/taskengine/worker_ctx_cancellation_test.go
```diff
@@ -0,0 +1,83 @@
...
+func TestRunBlockTriggerImmediately_ContextCancellationInterruptsWorkerCall(t *testing.T) {
+	// runBlockTriggerImmediately only reads n.logger (nil-guarded) and the
+	// global chain-state reader registry, so a zero-value Engine is enough —
+	// this keeps the test independent of the gateway config fixture.
+	engine := &Engine{}
+
+	const testChainID int64 = 987654321
+	reader := &blockingChainStateReader{started: make(chan struct{})}
+	RegisterChainStateReader(uint64(testChainID), reader)
+	defer RegisterChainStateReader(uint64(testChainID), nil)
```
chrisli30: This cleanup is a no-op. `RegisterChainStateReader` early-returns when `reader == nil` (chain_state_reader.go:635), so this `defer` never unregisters the fake — the blocking reader leaks into the global `chainStateReaderRegistry` after the test. It's harmless in practice (the chain id `987654321` is unique, so nothing collides), but it's misleading dead code.

The sibling tests use the real pattern instead:

```go
ClearChainStateReaderRegistry()
defer ClearChainStateReaderRegistry()
```

Suggest switching to that (see `chain_state_reader_test.go:395-396`).

---
_Generated by [Claude Code](https://claude.ai/code)_

core/taskengine/engine.go
```diff
@@ -3517,7 +3533,7 @@ func (n *Engine) SimulateWorkflow(user *model.User, trigger *avsproto.TaskTrigge
 	// Step 1: Start timing BEFORE trigger execution (consistent with node timing)
 	triggerStartTime := time.Now()
 
-	triggerOutput, err := n.runTriggerImmediately(triggerTypeStr, triggerConfig, inputVariables)
+	triggerOutput, err := n.runTriggerImmediately(context.Background(), triggerTypeStr, triggerConfig, inputVariables)
```
chrisli30: Coverage gap / inconsistency (non-blocking): `SimulateWorkflow` is also a request-context REST entry point (`aggregator/rest/handlers_workflows.go:336` → `s.engine.SimulateWorkflow(...)`), but this call keeps `context.Background()`. For a block-trigger simulation it reaches the same `GetBlockNumber` / `HeaderByNumber` worker calls (sites 5 & 6 in the issue), so request cancellation won't propagate on that path.

The issue's seven sites are technically satisfied via the `RunNode` / `RunTrigger` callers, so this isn't a missed acceptance criterion — but it leaves one entry point inconsistent. A small follow-up threading `ctx` into `SimulateWorkflow` would close it. Fine to defer.

---
_Generated by [Claude Code](https://claude.ai/code)_

chrisli30 (commented):
## Audit summary

Solid, well-scoped implementation of #599. Locally on the PR head: `go build ./...`, `go vet`, and lint pass, and the new test passes including under `-race` (×3). I have no blocking concerns — two non-blocking findings inline.

### CI failures are environment-related, not this PR
The four failing unit-test jobs (`core/taskengine`, `.../trigger`, `.../macros`, `erc4337/preset`) all failed due to key not set up from contributor’s branch. We will re-run the tests on staging branch.

### What's good
- All **seven** worker-call sites from the issue are threaded (`GetSmartWalletAddress` ×2, `FindMatchingWalletSalt` ×2, `GetBlockNumber` ×2, `HeaderByNumber`).
- The thin-wrapper / `...WithContext` sibling approach keeps exported signatures stable and avoids churning ~150 mostly-test call sites — a reasonable deviation from the issue's literal proposal, and well explained in the description.
- The async queue path (`Perform` → `RunTask`) correctly stays on `context.Background()`.
- Scoping to the four REST handlers is right: the aggregator-service gRPC handlers were deleted in the REST migration (`aggregator/rpc_server.go:500-506`), so REST is the only live request-context entry point.
- Error wrapping with `%w` is preserved, so `errors.Is(err, context.Canceled)` holds through `runBlockTriggerImmediately` — the test's assertion is valid.

### Findings (both non-blocking)
1. The test's cleanup `defer` is a no-op — see inline comment on `worker_ctx_cancellation_test.go`.
2. `SimulateWorkflow` REST path still uses `context.Background()` — see inline comment on `engine.go`.

Nit: double blank line after the `HeaderByNumber` stub in the new test file (cosmetic).

---
_Generated by [Claude Code](https://claude.ai/code)_

chrisli30:
Hi @Antrikshgwal, I’m following up on this PR. Could you take a look and address both copilot’s and mine comments?

Let me know if you have any feedback. Thanks.

Antrikshgwal:
> Hi @Antrikshgwal, I’m following up on this PR. Could you take a look and address both copilot’s and mine comments?
> 
> Let me know if you have any feedback. Thanks.

Hi @chrisli30, I made some changes. Please take a look.

chrisli30:
> > Hi @Antrikshgwal, I’m following up on this PR. Could you take a look and address both copilot’s and mine comments?
> > Let me know if you have any feedback. Thanks.
> 
> Hi @chrisli30, I made some changes. Please take a look.

Looking great, good to merge! Thanks for your contribution @Antrikshgwal

chrisli30 (approved):

PR #620, opened 2026-06-22
feat: ERC20 allowance overrides + user-facing erc20_overrides API for RunNodeImmediately

Antrikshgwal:
Fixes #413 

## Summary

Completes the remaining work on #413. The `SimulationStateMap` infra and the
balance-override path landed in #517; this PR adds the **allowance path** and the
**user-facing `erc20_overrides` API**, so a caller testing an isolated
`RunNodeImmediately` contract write (e.g. a Uniswap swap) can seed token balance
and approval up front instead of hitting `ERC20: transfer amount exceeds
allowance/balance`.

## Changes

- **proto** (`protobuf/avs.proto`): new `ERC20StateOverride` message and
  `repeated ERC20StateOverride erc20_overrides = 5` on `RunNodeWithInputsReq`.
  Regenerated `avs.pb.go` with the pinned toolchain (protoc v29.2 /
  protoc-gen-go v1.36.6) so the diff is limited to the new message + field.
- **REST/OpenAPI** (`api/openapi.yaml`, `aggregator/rest/...`): `ERC20StateOverride`
  schema + `erc20Overrides` on `RunNodeRequest`; regenerated types; mapping from
  the generated type to proto in the `RunNode` handler.
- **engine** (`core/taskengine/simulation_state.go`, `run_node_immediately.go`):
  - `ApplyUserERC20Override` — pure, absolute-value seeding of `balanceOf` /
    `allowance` storage slots (defaults: balance slot 0, allowance slot 3),
    plus a `parseUint256` hex/decimal helper.
  - Overrides are threaded from the request into the VM's `SimulationStateMap`
    and are **rejected in real-execution mode** — they only ever affect
    simulations, never deployed workflows or live wallets.
- **tests**: deterministic unit tests for the slot math, hex/decimal parsing,
  and validation (`simulation_state_override_test.go`); the Uniswap test now
  seeds overrides through the new path and skips cleanly without Tenderly config.
- **docs**: `TENDERLY_STATE_OVERRIDES.md` refreshed to document the shipped
  `SimulationStateMap` design and the new `erc20_overrides` API.

## Notes for reviewers

- **Field number**: used the next free tag `= 5` rather than the `= 10` in the
  original issue sketch (1/3/4 are used, 2 is reserved).
- **Uniswap test**: the issue suggested flipping it to "expect success", but its
  method (`QuoterV2.quoteExactInputSingle`) reverts *by design* to return the
  quote. The test now asserts the allowance/balance revert is gone (the actual
[...]

PR #624, opened 2026-06-23
feat: user-facing erc20_overrides for RunNodeImmediately (+ simulation-state fix & RPC E2E proof)

chrisli30:
Continues #620 (by @Antrikshgwal) and gets its E2E coverage green. Based on that branch, with two additions: the missing wiring fix that makes `erc20_overrides` actually work through the RPC path, and a decisive end-to-end test that proves it.

## What #620 was missing

@Antrikshgwal wired `erc20_overrides` from the request → proto → `RunNodeImmediately` → `SimulationStateMap`. Driving that path for real (not via a direct `ApplyUserERC20Override` call, per @will-dz's review) surfaced a bug the helper-only tests couldn't: **the overrides never reached Tenderly.**

`VM.RunNodeWithInputs` builds a fresh `tempVM` for isolated single-node execution but never copied the parent VM's `simulationState`. The seeded balances/allowances were applied to the parent VM and then dropped — so the contractWrite node simulated against empty state and the Uniswap swap still reverted with `ERC20: transfer amount exceeds allowance`. That's why the feature couldn't be shown green.

**Fix:** propagate `v.simulationState` into `tempVM` (`core/taskengine/vm.go`). When simulation state is unset this is `nil` (unchanged); when set-but-empty the Tenderly client still takes its default `state_objects` path, so existing tests are unaffected.

## The E2E test (addresses the review checklist)

`TestRunNodeImmediatelyRPC/ERC20Overrides_UniswapSwap` now drives the real `RunNodeImmediatelyRPC` entrypoint with a populated `erc20_overrides` on a Uniswap `exactInputSingle` contractWrite node, and asserts a true before/after:

- **control** (no overrides) reverts on the allowance/balance precondition, and
- the **same swap succeeds** once overrides seed USDC balance (slot 9) + SwapRouter02 allowance (slot 10).

Only balance/allowance flowing from the request all the way into the Tenderly simulation can flip that outcome. Slots 9/10 were confirmed empirically against the deployed Sepolia USDC (Circle FiatToken).

## Local verification (real Tenderly + funded `OWNER_EOA`)

Full `core/taskengine` package under `go test . -short` (the exact CI invocation):

| Test | baseline (#620 head) | this branch |
| --- | --- | --- |
| `TestRunNodeImmediatelyRPC` (incl. `ERC20Overrides_UniswapSwap`) | ❌ FAIL | ✅ PASS |
| `TestContractWriteNode_UniswapV3Quote` | ✅ | ✅ |
| `TestExecuteTask_UniswapApprovalAndSwap_DifferentApprovalAmounts_Sepolia` | ✅ | ✅ |
| `TestContractWriteTenderlySimulation` | ✅ | ✅ |

The only remaining `-short` failure is `TestBalanceNode_MissingAPIKey`, which fails **identically on the #620 baseline** (pre-existing, unrelated to this change) and **skips in CI** when `MORALIS_API_KEY` is configured — consistent with the currently-green `staging` runs.

### DIFF
--- a/aggregator/rest/generated/types.gen.go
+++ b/aggregator/rest/generated/types.gen.go
@@ -471,7 +471,7 @@ type ERC20StateOverride struct {
 	// Allowance Allowance override (hex 0x… or decimal string).
 	Allowance *string `json:"allowance,omitempty"`
 
-	// AllowanceSlot Storage slot for the allowance mapping (default: 3).
+	// AllowanceSlot Storage slot for the allowance mapping (default: 1).
 	AllowanceSlot *int64 `json:"allowanceSlot,omitempty"`
 
 	// Balance Balance override (hex 0x… or decimal string).
--- a/aggregator/rest/generated/types.gen.go
+++ b/aggregator/rest/generated/types.gen.go
@@ -466,6 +466,30 @@ type CustomCodeNodeConfig struct {
 	Source string `json:"source"`
 }
 
+// ERC20StateOverride Seeds a token's balanceOf / allowance storage slots for a single simulation. balanceOf[owner] lives at keccak256(abi.encode(owner, balanceSlot)); allowance[owner][spender] at keccak256(abi.encode(spender, keccak256(abi.encode(owner, allowanceSlot)))).
+type ERC20StateOverride struct {
+	// Allowance Allowance override (hex 0x… or decimal string).
+	Allowance *string `json:"allowance,omitempty"`
+
+	// AllowanceSlot Storage slot for the allowance mapping (default: 3).
+	AllowanceSlot *int64 `json:"allowanceSlot,omitempty"`
+
+	// Balance Balance override (hex 0x… or decimal string).
+	Balance *string `json:"balance,omitempty"`
+
+	// BalanceSlot Storage slot for the balanceOf mapping (default: 0).
+	BalanceSlot *int64 `json:"balanceSlot,omitempty"`
+
+	// OwnerAddress Lowercase or checksummed hex EOA / contract address.
+	OwnerAddress EthereumAddress `json:"ownerAddress"`
+
+	// SpenderAddress Lowercase or checksummed hex EOA / contract address.
+	SpenderAddress *EthereumAddress `json:"spenderAddress,omitempty"`
+
+	// TokenAddress Lowercase or checksummed hex EOA / contract address.
+	TokenAddress EthereumAddress `json:"tokenAddress"`
+}
+
 // ETHTransferNode defines model for ETHTransferNode.
 type ETHTransferNode struct {
 	Config *ETHTransferNodeConfig `json:"config,omitempty"`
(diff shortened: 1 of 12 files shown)


chrisli30:
Thanks for the review — addressed in 87e7db9:

**Default allowance slot 3 → 1.** Correct catch: slot 3 is `_name` in the standard OpenZeppelin ERC20 layout, so omitting `allowance_slot` would silently seed the wrong slot. Defaulted to **1** (OZ keeps `_balances` at slot 0, `_allowances` at slot 1) and made it consistent across the engine const, proto, OpenAPI (+ regenerated types), the docs (which also self-contradicted), and the uniswap test's candidate slots.

**Negative slot coercion.** The handler now treats a negative slot as *unset* (engine applies its default) instead of silently coercing to slot 0, and the OpenAPI schema documents `minimum: 0` for both slot fields.

**salt:0 control nondeterminism.** This was flagged against an earlier revision. The override E2E test was since reworked to use a **deterministic, never-funded** owner (`0xe8a78b47…`), so the counterfactual salt:0 runner is guaranteed to hold 0 USDC / 0 allowance — the control reverts reliably and the before/after stays meaningful. It also now isolates the balance path explicitly: `no overrides → exceeds allowance`, `allowance-only → exceeds balance`, `balance+allowance → success`.

chrisli30:
Round 2 addressed in 279cabb:
- **Magnitude comment:** correct — `0x38d7ea4c68000` is 1e15 base units = **1,000,000,000 USDC** at 6 decimals, not 1,000,000. Fixed the comment in all five spots (it was an oversized-on-purpose balance; only the comment was wrong).
- **Max-uint256 string concat:** replaced with the full hex literal in the unit test.
- Also dropped the misleading `Standard ERC20 | 0 | 3` row from the slot table (OpenZeppelin v4/v5 `0/1` is the real standard), consistent with the require-explicit-slot change.

chrisli30:
Round 3 addressed in the latest commit: corrected the proto/OpenAPI/generated docstrings — `erc20_overrides` are **rejected with an error** when `isSimulated=false` (`erc20_overrides are only supported in simulation mode`), not silently ignored. Thanks!

chrisli30:
Round 4 addressed in the latest commit:
- **Silent type-assertion drop:** `applyERC20Overrides` now returns an explicit internal error if the stashed value isn't a `[]*ERC20StateOverride` (it could only happen via a wiring bug), instead of silently dropping the overrides. Empty slice still no-ops.
- **QuoterV2 override mismatch:** you're right — `quoteExactInputSingle` reverts by design to return the quote and never does a `transferFrom`, so seeding a SwapRouter02 allowance proved nothing there. Removed the override seeding; that test now honestly validates tuple-parameter calldata generation, and the override request path is proven E2E by `TestRunNodeImmediatelyRPC/ERC20Overrides_UniswapSwap`.

chrisli30:
Round 5 addressed in the latest commit: the override E2E test now asserts both the `StoreWallet` error and the `structpb.NewValue` ABI-build error (`require.NoError`) so a seed/ABI failure fails the test deterministically rather than surfacing later as a confusing error.

chrisli30:
Round 6 addressed in the latest commit: the "allowance only" phase now accepts either the inner `transfer amount exceeds balance` revert or TransferHelper `STF`, since Tenderly may surface SwapRouter02's transferFrom failure as either — removing the brittleness without weakening the proof (it still asserts the allowance was seeded and the remaining failure is the balance precondition).

PR #638, opened 2026-06-26
chore: make dev-stack/dev-gateway targets mirror studio start.sh

chrisli30:
The `dev-gateway` / `dev-worker-*` / `dev-stack` make targets referenced `config/gateway.yaml` + `config/worker-*.yaml`, which don't exist in the studio local-dev flow — that uses the `-dev.yaml` configs (symlinked from the env repo) that `studio/scripts/start.sh` runs. This points the targets at the same configs and sources `.env.local` so the `${SEPOLIA_BUNDLER_URL}`/`${BASE_SEPOLIA_BUNDLER_URL}` refs resolve; `dev-stack` fails fast if they're unset.

`make dev-stack` = gateway (:8080 REST / :2206 gRPC) + sepolia worker (:50051) + base-sepolia worker (:50052) + sepolia operator, all torn down together on Ctrl-C.

Note: depends on the `-dev.yaml` configs being present (they're local symlinks into the env repo, same as start.sh already assumes).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -276,21 +276,21 @@ dev-gateway: build
 	@mkdir -p logs
 	@echo "🚀 Starting gateway (dev) — REST :8080, gRPC :2206"
 	@echo "📝 Logs: logs/gateway.log"
-	./out/ap aggregator --config=config/gateway.yaml 2>&1 | tee logs/gateway.log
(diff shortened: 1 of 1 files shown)


chrisli30:
Superseding directly on staging: dropping the `-dev` suffix from the config filenames (so the Makefile's original `config/gateway.yaml` paths are correct as-is) plus a build-free `run-*` single-source-of-truth refactor shared by the Makefile and studio start.sh.

PR #648, opened 2026-06-28
fix: surface WAITING execution status over REST (was masked as pending)

chrisli30:
**Draft** — closes a gap found in review of the durable-execution SDK hand-off.

## The gap
A paused (durable-execution) execution is `EXECUTION_STATUS_WAITING` in the engine, but the REST layer never mapped it:
- `executionStatusProtoToWire` had no `WAITING` case → fell through to `"pending"`.
- The OpenAPI `ExecutionStatus` enum omitted it (`pending | success | failed | error`).

**Net effect:** a client couldn't distinguish "awaiting approval / chain event" from an ordinary in-flight `"pending"` execution. The feature still worked (polling helpers keep waiting either way), but the `WAITING` state was invisible to consumers.

## Fix
- OpenAPI `ExecutionStatus` enum gains `"waiting"` (+ description); `make rest-gen` → `generated.ExecutionStatusWaiting`.
- `executionStatusProtoToWire` maps `EXECUTION_STATUS_WAITING` → `"waiting"`.
- Test pins the full proto→wire vocabulary incl. `WAITING`, with an enum-drift guard against the generated constant.
- SDK hand-off doc updated: the lifecycle section now states the wire value is `"waiting"` and notes this fix.

storage-check clean (enum addition is additive); aggregator suite green.

### DIFF
--- a/SDK_HANDOFF_DURABLE_EXECUTION.md
+++ b/SDK_HANDOFF_DURABLE_EXECUTION.md
@@ -67,12 +67,14 @@ Chain-event awaits are **not** resumed through this endpoint — an operator fir
 
 ---
 
-## The `WAITING` lifecycle (important for polling/streaming)
+## The `waiting` lifecycle (important for polling/streaming)
 
-`EXECUTION_STATUS_WAITING` is **non-terminal**. Any SDK helper that "waits for an execution to finish" must treat `WAITING` as *keep waiting*, not done:
+Over REST the status is the lowercase wire value **`"waiting"`** (the generated `ExecutionStatus` union member; proto enum `EXECUTION_STATUS_WAITING`). It is **non-terminal**. Any SDK helper that "waits for an execution to finish" must treat `"waiting"` as *keep waiting*, not done:
 
-- `executions.get` / the SSE stream will report `WAITING` while paused, then a terminal status (`SUCCESS` / `FAILED` / `ERROR`) once resumed (or `FAILED` with "await timed out…" if the timeout sweep fires).
-- A `waitForExecution`-style util that currently stops on "any status set" or "not PENDING" will return early on `WAITING` — update its terminal-status predicate to exclude `WAITING`.
+- `executions.get` / the SSE stream report `"waiting"` while paused, then a terminal status (`"success"` / `"failed"` / `"error"`) once resumed (or `"failed"` with "await timed out…" if the timeout sweep fires).
+- A `waitForExecution`-style util that currently stops on "any status set" or "not `pending`" will return early on `"waiting"` — update its terminal-status predicate to exclude `"waiting"`.
+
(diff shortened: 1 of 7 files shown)


chrisli30:
## Audit: correct fix, but **incomplete** — it patches 1 of 3 status mappers

The fix to `executionStatusProtoToWire` (`aggregator/rest/mapping/execution.go:242`), the OpenAPI enum + regenerated `ExecutionStatusWaiting`, and the drift-guarded test are clean. That mapper powers `ProtoToOpenAPIExecution`, so **`GET /executions/{id}` now correctly reports `"waiting"`.** 👍

But the proto→wire status switch is **duplicated in three places**, each with the same `default: return "pending"` that masks WAITING — and only one was patched:

| Function | Location | Powers | After this PR |
|---|---|---|---|
| `executionStatusProtoToWire` | `mapping/execution.go:240` | `GET /executions/{id}` | ✅ fixed |
| `executionStatusToWire` | `handlers_executions.go:306` | **`:getStatus`** (`:166`) + **`:stream`** SSE (`:263`) | ❌ still `"pending"` |
| `protoExecStatusToOpenAPI` | `handlers_workflows.go:598` | **`:trigger`** response (`:258`) | ❌ still `"pending"` |

So after this PR:
- **`GET /executions/{id}:getStatus`** → a paused execution still returns `"pending"`.
- **`GET /executions/{id}:stream`** (SSE) → still emits `"pending"`, never a distinct `"waiting"` frame (`handlers_executions.go:263` calls the unpatched `executionStatusToWire`).
- **`:trigger`** response → still `"pending"`.

These are exactly the lightweight polling paths a client uses to watch for resume — and this PR's own hand-off edit now claims *"`executions.get` / the SSE stream report `"waiting"` while paused,"* which is true for `get` but **false for the stream and getStatus**.

### Root cause of the miss
The doc comment at `mapping/execution.go:88-90` says `ProtoExecutionToOpenAPISummary` "is returned by the GetExecutionStatus and StreamExecution endpoints." It isn't — both handlers use their own local `executionStatusToWire` and bypass the mapper. The fix looks complete if you trust that comment.

### Recommendation
1. Add the `WAITING → "waiting"` case to **`executionStatusToWire`** and **`protoExecStatusToOpenAPI`** too (minimum to ship).
2. Better: **collapse all three into the single mapper** so this can't drift again — the handler-local copies exist for file-locality, not behavior.
3. Extend the test to cover the surviving function(s).
4. Fix the stale `:88-90` comment.

Functional note: an SDK `waitForTerminal` that uses a terminal *allow-list* (`success`/`failed`/`error`) keeps working regardless — but the *observability* goal here (let a client show "awaiting approval") only lands on `GET /executions/{id}`, not on getStatus/stream.

> 🤖 Audit via Claude Code (ava-sdk-js durable-execution hand-off review)

PR #657, opened 2026-07-05
fix: parseUnit macro implements ethers parseUnits scaling with fractional support

### LINKED ISSUE #656: fix: parseUnit macro divides by decimal count and rejects fractional input
### Summary

The `parseUnit` expression builtin divides the value by the **raw decimal count** instead of scaling by `10^decimals`, and it cannot accept a fractional value at all. Both defects make it unusable for its documented purpose (building a fixed-point price threshold), and any task condition using it compares against a meaningless number.

### Location

[`core/taskengine/macros/exp.go:118`](https://github.com/AvaProtocol/EigenLayer-AVS/blob/staging/core/taskengine/macros/exp.go#L118)

```go
func ParseUnit(val string, decimal uint) *big.Int {
	b, ok := ethmath.ParseBig256(val)
	if !ok {
		panic(fmt.Errorf("Parse error: %s", val))
	}

	r := big.NewInt(0)
	return r.Div(b, big.NewInt(int64(decimal)))   // <-- divides by `decimal` (e.g. 8), not 10^decimal
}
```

Registered as the user-facing `parseUnit` builtin at [`exp.go:206`](https://github.com/AvaProtocol/EigenLayer-AVS/blob/staging/core/taskengine/macros/exp.go#L206).

### Defects

**1. Wrong scaling.** The name mirrors ethers.js [`parseUnits(value, decimals)`](https://docs.ethers.org/v6/api/utils/#parseUnits), the universal EVM convention, which returns `value * 10^decimals`. The implementation instead computes `value / decimals` (integer division by the raw count).

| Call | Current result | Expected (`value * 10^decimals`) |
|------|----------------|-----------------------------------|
| `parseUnit("2621", 8)`  | `327` (2621 / 8) | `262100000000` |
| `parseUnit("100", 6)`   | `16` (100 / 6)   | `100000000` |

**2. No fractional input.** `ethmath.ParseBig256` only parses integers/hex, so `parseUnit("2621.99", 8)` panics (`Parse error: 2621.99`). Price thresholds are inherently fractional, so this is exactly the input the builtin needs to accept.

### Impact

`parseUnit` is exposed to users in task-condition expressions (documented use: a price threshold compared against `chainlinkPrice(...)`, which returns an 8-decimal integer). Because the produced threshold is wrong by orders of magnitude — or panics outright on a fractional value — conditions fire when they shouldn't or never fire. Silent, incorrect automation behavior with no error surfaced.

### Proposed fix

Reimplement `ParseUnit` to match ethers `parseUnits`:
- scale by `10^decimals`
- accept a fractional component, right-padded to `decimals`
- reject over-precision (fraction longer than `decimals`), negatives, and non-numeric input

```go
func ParseUnit(val string, decimal uint) *big.Int {
	parts := strings.SplitN(val, ".", 2)
	whole, ok := new(big.Int).SetString(parts[0], 10)
	// ... scale whole by 10^decimals, add right-padded fraction, reject over-precision/negatives
}
```

No existing tests or callers depend on the current (broken) output — `parseUnit` has zero coverage and is only referenced in a commented-out example — so the fix is safe for anyone reasonably expecting ethers semantics.

Antrikshgwal:
## What

Fix the `parseUnit` expression builtin, which computed `value / decimals` instead
of scaling by `10^decimals`. Reimplemented to match the ethers.js
`parseUnits(value, decimals)` convention, including fractional input.

Closes #656.

## Why

`parseUnit` is exposed to users in task-condition expressions — the documented use
is building a fixed-point price threshold to compare against `chainlinkPrice(...)`,
which returns an 8-decimal integer. The old implementation divided by the raw
decimal *count*:

```go
r.Div(b, big.NewInt(int64(decimal)))   // parseUnit("2621", 8) -> 327
```

So every threshold was wrong by orders of magnitude (and lossy, from integer
division). Conditions fired when they shouldn't or never fired — silent, incorrect
automation with no error surfaced. On top of that, the function couldn't accept a
fractional value like `"2621.99"` at all (it used `ParseBig256`, integer-only), so
it was unusable for the price thresholds it exists to build.

## What changed

`ParseUnit` now:
- scales by `10^decimals` (`parseUnit("2621", 8)` → `262100000000`)
- accepts a fractional component, right-padded to `decimals` (`parseUnit("2621.99", 8)` → `262199000000`)
- rejects over-precision (fraction longer than `decimals`), negatives, and non-numeric input

| Call | Before | After |
|------|--------|-------|
| `parseUnit("2621", 8)`    | `327`      | `262100000000` |
| `parseUnit("2621.99", 8)` | panic      | `262199000000` |
| `parseUnit("100", 6)`     | `16`       | `100000000` |

## Tests

Added `TestParseUnit` (pure unit test, no RPC): the regression assertion
(`2621.99 → 262199000000`), integer scaling, and the over-precision rejection.

## Compatibility

No callers or tests depended on the old output — the only reference was a
commented-out example. Behavior now matches the ethers `parseUnits` semantics the
name implies.

### DIFF
--- a/core/taskengine/macros/exp.go
+++ b/core/taskengine/macros/exp.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"math/big"
+	"regexp"
 	"strings"
 
 	"github.com/ethereum/go-ethereum/accounts/abi"
@@ -59,7 +60,7 @@ func readContractData(contractAddress string, data string, method string, contra
 }
 
 // QueryContract
-//const taskCondition = `cmp(chainlinkPrice("0x694AA1769357215DE4FAC081bf1f309aDC325306"), parseUnit("262199799820", 8)) > 1`
+//const taskCondition = `cmp(chainlinkPrice("0x694AA1769357215DE4FAC081bf1f309aDC325306"), parseUnit("2621.99", 8)) > 1`
 
 func chainlinkLatestRoundData(tokenPair string) *big.Int {
 	output, err := QueryContract(
@@ -115,22 +116,35 @@ func BigLt(a *big.Int, b *big.Int) bool {
 	return a.Cmp(b) < 0
 }
 
-// ParseUnit converts a decimal string into a fixed-point integer scaled by
-// 10^decimals, matching the ethers.js parseUnits(value, decimals) convention.
-// It accepts a fractional component (e.g. ParseUnit("2621.99", 8) == 262199000000),
-// which is what task-condition expressions need to build a price threshold to
-// compare against chainlinkPrice(). A fraction with more digits than `decimals`
-// or a negative value is rejected.
-func ParseUnit(val string, decimal uint) *big.Int {
-	parts := strings.SplitN(val, ".", 2)
+// parseUnitRe matches a non-negative decimal number with an optional fractional
+// part (e.g. "2621", "2621.99", "2621."). Validating the shape up front rejects
+// signs, a leading "+" in the fraction, hex, and other malformed input that
+// big.Int.SetString would otherwise silently accept. A bare ".5" (empty whole
+// part) is intentionally rejected; task thresholds are written with a leading 0.
+var parseUnitRe = regexp.MustCompile(`^[0-9]+(\.[0-9]*)?$`)
 
-	whole, ok := new(big.Int).SetString(parts[0], 10)
-	if !ok {
-		panic(fmt.Errorf("Parse error: %s", val))
+// parseUnitMaxDecimals caps 10^decimals so a user-supplied `decimal` can't force
+// the aggregator/operator to materialize a huge integer. 10^78 ≈ 2^259, beyond
+// anything on-chain needs (uint256 tops out near 10^77).
+const parseUnitMaxDecimals = 78
+
+// ParseUnit converts a non-negative decimal string into a fixed-point integer
+// scaled by 10^decimals, matching the ethers.js parseUnits(value, decimals)
+// convention. It accepts a fractional component (e.g. ParseUnit("2621.99", 8) ==
+// 262199000000), which is what task-condition expressions need to build a price
+// threshold to compare against chainlinkPrice(). Input that is not a non-negative
+// decimal, a fraction with more digits than `decimals` (over-precision), or
+// `decimals` beyond the on-chain range is rejected.
+func ParseUnit(val string, decimal uint) *big.Int {
+	if !parseUnitRe.MatchString(val) {
+		panic(fmt.Errorf("Parse error: %q is not a non-negative decimal number", val))
 	}
-	if whole.Sign() < 0 {
-		panic(fmt.Errorf("Parse error: negative value not supported: %s", val))
+	if decimal > parseUnitMaxDecimals {
+		panic(fmt.Errorf("Parse error: decimals %d out of range (max %d)", decimal, parseUnitMaxDecimals))
 	}
+
+	parts := strings.SplitN(val, ".", 2)
+	whole, _ := new(big.Int).SetString(parts[0], 10) // regex guarantees a valid non-negative integer
 
 	scale := new(big.Int).Exp(big.NewInt(10), new(big.Int).SetUint64(uint64(decimal)), nil)
 	result := new(big.Int).Mul(whole, scale)
@@ -141,10 +155,7 @@ func ParseUnit(val string, decimal uint) *big.Int {
 		if uint(len(frac)) > decimal {
 			panic(fmt.Errorf("Parse error: fractional part %q exceeds %d decimals in %q", frac, decimal, val))
 		}
-		fracVal, ok := new(big.Int).SetString(frac+strings.Repeat("0", int(decimal)-len(frac)), 10)
-		if !ok || fracVal.Sign() < 0 {
-			panic(fmt.Errorf("Parse error: %s", val))
-		}
+		fracVal, _ := new(big.Int).SetString(frac+strings.Repeat("0", int(decimal)-len(frac)), 10)
 		result.Add(result, fracVal)
 	}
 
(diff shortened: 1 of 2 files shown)


will-dz:
Thanks for this fix, @Antrikshgwal. The core change looks correct and well-motivated. The old implementation divided by the decimal *count* instead of scaling by `10^decimals`, and the new behavior matches the ethers `parseUnits` semantics the name promises. The regression test is a good addition too (verified it passes locally).

Before merging, three issues in the input validation — the first two confirmed by running the new code:

**1. `parseUnit("-0.5", 8)` returns positive `50000000` (bug)**

The negative check uses `whole.Sign() < 0`, but for `-0.5` the whole part parses as `-0`, whose `Sign()` is `0`, so the check is bypassed and the sign is silently dropped. A silently sign-flipped threshold is exactly the class of silent-wrong-comparison bug this PR sets out to fix. (`-1.5` is correctly rejected; only the `-0.x` family slips through.)

**2. `parseUnit("1.+5", 8)` returns `105000000` (bug)**

`big.Int.SetString` accepts a leading `+`, so a malformed fraction like `1.+5` is accepted as `1.05` instead of being rejected.

Both are fixed by validating the input shape up front instead of relying on `SetString`:

    var parseUnitRe = regexp.MustCompile(`^[0-9]+(\.[0-9]*)?$`)

    func ParseUnit(val string, decimal uint) *big.Int {
        if !parseUnitRe.MatchString(val) {
            panic(fmt.Errorf("Parse error: %q is not a non-negative decimal number", val))
        }
        ...
    }

This also lets you drop the `whole.Sign()` / `fracVal.Sign()` checks entirely. Please add test cases for `-0.5`, `-1.5`, and a malformed fraction.

**3. Unbounded `10^decimals` (resource exhaustion)**

`decimal` comes from user-submitted task conditions evaluated on the aggregator/operator, and `Exp(big.NewInt(10), decimal, nil)` is unbounded — `parseUnit("1", 4000000000)` attempts to materialize a ~1.7 GB integer. The expr VM recovers panics, but it can't recover an OOM. The old code didn't have this exposure (it only divided). A cap fixes it:

    if decimal > 78 { // 10^78 ≈ 2^259; nothing on-chain needs more
        panic(fmt.Errorf("Parse error: decimals %d out of range", decimal))
    }

**Minor / non-blocking**

- The commented-out example on line 62 (`parseUnit("262199799820", 8)`) now means something different under the new semantics — worth updating to `parseUnit("2621.99", 8)` while you're here.
- `.5` is rejected (ethers accepts it, treating the empty whole part as 0). Fine to leave as-is, but worth a conscious decision since the PR advertises ethers compatibility.
- Hex input (`0x...`) was accepted by the old `ParseBig256` path and no longer is — given the semantics change and no existing callers, this seems right, just noting it.

**CI:** the failing jobs (`macros`, `taskengine`, `trigger`, `erc4337/preset`) **is unrelated to this diff** — those packages contain tests that hit live Sepolia RPC, and fork PRs don't get the repo's RPC secrets. The new `TestParseUnit` passes.

Antrikshgwal:
Thanks for the review @will-dz. 
All three confirmed issues are real. Pushed a fix addressing them.

Antrikshgwal:
> **1. `parseUnit("-0.5", 8)` returns positive `50000000` (bug)**
> 
> The negative check uses `whole.Sign() < 0`, but for `-0.5` the whole part parses as `-0`, whose `Sign()` is `0`, so the check is bypassed and the sign is silently dropped. A silently sign-flipped threshold is exactly the class of silent-wrong-comparison bug this PR sets out to fix. (`-1.5` is correctly rejected; only the `-0.x` family slips through.)
> 
> **2. `parseUnit("1.+5", 8)` returns `105000000` (bug)**
> 
> `big.Int.SetString` accepts a leading `+`, so a malformed fraction like `1.+5` is accepted as `1.05` instead of being rejected.
> 
> Both are fixed by validating the input shape up front instead of relying on `SetString`:
> 
> ```
> var parseUnitRe = regexp.MustCompile(`^[0-9]+(\.[0-9]*)?$`)
> 
> func ParseUnit(val string, decimal uint) *big.Int {
>     if !parseUnitRe.MatchString(val) {
>         panic(fmt.Errorf("Parse error: %q is not a non-negative decimal number", val))
>     }
>     ...
> }
> ```
> 
> This also lets you drop the `whole.Sign()` / `fracVal.Sign()` checks entirely. Please add test cases for `-0.5`, `-1.5`, and a malformed fraction.
> 
switched to up-front shape validation instead of relying on `SetString`:

`var parseUnitRe = regexp.MustCompile(`^[0-9]+(\.[0-9]*)?$`)`

Any leading sign (-, and the + inside the fraction) is now rejected before parsing.

Antrikshgwal:
> **3. Unbounded `10^decimals` (resource exhaustion)**
> 
> `decimal` comes from user-submitted task conditions evaluated on the aggregator/operator, and `Exp(big.NewInt(10), decimal, nil)` is unbounded — `parseUnit("1", 4000000000)` attempts to materialize a ~1.7 GB integer. The expr VM recovers panics, but it can't recover an OOM. The old code didn't have this exposure (it only divided). A cap fixes it:
> 
> ```
> if decimal > 78 { // 10^78 ≈ 2^259; nothing on-chain needs more
>     panic(fmt.Errorf("Parse error: decimals %d out of range", decimal))
> }
> ```
capped decimal before the Exp:

`const parseUnitMaxDecimals = 78 // 10^78 ≈ 2^259; uint256 tops out near 10^77`

Also added -0.5, -1.5, the malformed fraction, plus an out-of-range decimals test cases.

will-dz:
Verified the follow-up commit (2aa0060), and all three issues are resolved:

- `-0.5` / `-1.5` / `1.+5` / hex / empty input are now rejected by the up-front shape validation; the sign-drop bug is gone.
- `decimals` is capped at 78 before the `Exp` call, so the resource-exhaustion path is closed.
- Test table covers all the cases requested (negatives, malformed fraction, over-precision, out-of-range decimals), and the stale example comment was updated too.

Ran `TestParseUnit` and re-checked the edge cases locally against this commit — all behave correctly, and happy paths are unchanged (`007.50` → `750000000`, `0.00000001` → `1`).

LGTM 🎉 Thanks for the quick turnaround, @Antrikshgwal!

PR #668, opened 2026-07-25
fix: validate chainlink feed freshness in price macros

### LINKED ISSUE #661: fix: chainlink price macros ignore feed staleness and negative answers
### Summary

The Chainlink price macros return the feed's `answer` while discarding every field that says whether that answer is still valid. If a feed stalls or goes down, price-conditioned tasks keep evaluating against a frozen price and fire (or fail to fire) on data that no longer reflects the market. The code already flags this at [`exp.go:79`](https://github.com/AvaProtocol/EigenLayer-AVS/blob/staging/core/taskengine/macros/exp.go#L79).

### Location

[`core/taskengine/macros/exp.go:65-104`](https://github.com/AvaProtocol/EigenLayer-AVS/blob/staging/core/taskengine/macros/exp.go#L65-L104)

```go
func chainlinkLatestRoundData(tokenPair string) *big.Int {
	output, err := QueryContract(rpcConn, common.HexToAddress(tokenPair), chainlinkABI, "latestRoundData")
	if err != nil {
		panic(...)
	}

	// TODO: Check round and answer to prevent the case where chainlink down we
	// may got outdated data
	if len(output) < 2 || output[1] == nil {
		return big.NewInt(0)
	}
	return output[1].(*big.Int)   // answer only
}
```

Exposed to users as the `priceChainlink`, `chainlinkPrice`, and `latestRoundDataChainlink` builtins ([`exp.go:210-212`](https://github.com/AvaProtocol/EigenLayer-AVS/blob/staging/core/taskengine/macros/exp.go#L210-L212)).

### Defects

**1. Staleness is never checked.** The bundled `EACAggregatorProxy` ABI returns:

```
latestRoundData() -> (roundId uint80, answer int256, startedAt uint256, updatedAt uint256, answeredInRound uint80)
```

The implementation reads `output[1]` (`answer`) and drops `updatedAt` (`output[3]`), `roundId` (`output[0]`), and `answeredInRound` (`output[4]`) — exactly the fields that reveal a stalled feed. The canonical Chainlink consumer checks are `updatedAt` against a max-age window and `answeredInRound >= roundId`; neither happens here.

**2. Negative answers are not guarded.** `answer` is `int256`. A negative or zero value (feed misconfiguration, or a genuinely negative feed) is passed straight into `bigCmp`/`bigGt`/`bigLt` as a valid price.

**3. `latestAnswer` is deprecated.** `priceChainlink` and `chainlinkPrice` — the two most likely builtins for a user to reach for — call `latestAnswer` ([`exp.go:94`](https://github.com/AvaProtocol/EigenLayer-AVS/blob/staging/core/taskengine/macros/exp.go#L94)), which Chainlink deprecated *precisely because* it returns no round or timestamp data. There is no way to detect staleness through that path at all.

**4. A failed round silently reads as price 0.** When the output is short or nil, the function returns `big.NewInt(0)` rather than failing. A threshold like `bigLt(chainlinkPrice(feed), parseUnit("2000", 8))` evaluates **true** on a dead feed.

### Impact

These macros exist to gate task execution on price. A stale, zero, or negative price produces a silently wrong comparison — the task fires when it shouldn't, or stays dormant when it should fire, with no error surfaced to the user. Same class of failure as #656, but triggered by live market/feed conditions rather than a mis-set threshold.

### Design question 

`chainlinkLatestRoundData` returns a bare `*big.Int`, so there is nowhere to express a staleness window or an error. Options:

1. **Extra macro param** — `latestRoundDataChainlink(feed, maxAgeSeconds)`. Explicit and per-task, but changes the builtin's signature.
2. **Config-level default max age** on the aggregator/operator, applied to all feed reads. No expression changes; one global policy.
3. **Panic on stale** — the expr VM already recovers panics (`TestExpressionPanicWonCrash`), so the task fails loudly instead of evaluating on bad data. Fits the existing style of these functions, which already panic on RPC error.

Related: should `priceChainlink`/`chainlinkPrice` be re-pointed at `latestRoundData` (so they can be staleness-checked at all), keeping the names but changing the underlying call? That is a user-visible behavior change and needs a maintainer call.

### Proposed fix

Pending the decision above, roughly:

```go
// validate the round before trusting the answer
if answeredInRound.Cmp(roundId) < 0 { /* stale round */ }
if updatedAt is older than maxAge  { /* stalled feed */ }
if answer.Sign() <= 0              { /* invalid price */ }
```

Happy to implement whichever direction is preferred — flagging the API question first rather than assuming.

Antrikshgwal:
Closes #661.

The chainlink price macros returned the feed `answer` while discarding every
field that signals validity, so a stalled or down feed let price-conditioned
tasks evaluate against a frozen price. A failed round also read as `0`, and
`priceChainlink`/`chainlinkPrice` used the deprecated `latestAnswer`, which
carries no timestamp to check at all.

**Changes**
- Extract `validateChainlinkRound` (unit-testable without a live feed): rejects a
  stale (`updatedAt` older than max age), incomplete, carried-over, or
  non-positive round.
- Panic on an invalid round instead of returning a silent `0`, so the expr VM
  fails the task loudly rather than acting on bad data.
- Re-point `priceChainlink`/`chainlinkPrice` at `latestRoundData` and drop the
  deprecated `latestAnswer` path.
- Add an optional `maxAgeSeconds` argument to override the default 24h window:
  `latestRoundDataChainlink(feed)` or `latestRoundDataChainlink(feed, 3600)`.

**Decisions worth a look**
- Default max age is a loose 24h so healthy slow-heartbeat feeds aren't rejected;
  callers on faster feeds should pass an explicit window.
- The `latestAnswer` re-point is a behavior change flagged in #661 — same return
  value, but these now panic on a stale feed instead of silently returning a price.

**Tests:** `validateChainlinkRound` unit tests (6 rejection cases + fresh passes)
and `chainlinkMaxAge` default/override, all without RPC.

---

**⚠️ User-visible behavior change (SDK consumers)**

`priceChainlink` / `chainlinkPrice` previously returned a price unconditionally. They now route through `latestRoundData` and **panic on a round stale beyond the max age** (default 24h), which the expr VM surfaces as a task failure. A task using these on a feed stale >24h will now fail loudly instead of evaluating against a frozen price. Pass `priceChainlink(feed, maxAgeSeconds)` to tune the window.

### DIFF
--- a/core/taskengine/macros/chainlink_test.go
+++ b/core/taskengine/macros/chainlink_test.go
@@ -1,0 +1,58 @@
+package macros
+
+import (
+	"math/big"
+	"testing"
+	"time"
+)
+
+func TestValidateChainlinkRound(t *testing.T) {
+	now := time.Unix(1_700_000_000, 0)
+	maxAge := time.Hour
+	tenMinAgo := big.NewInt(now.Add(-10 * time.Minute).Unix())
+	twoHoursAgo := big.NewInt(now.Add(-2 * time.Hour).Unix())
+	price := big.NewInt(262199000000)
+
+	// Fresh, complete round: the answer passes through unchanged.
+	got, err := validateChainlinkRound(chainlinkRound{
+		roundID: big.NewInt(100), answer: price, updatedAt: tenMinAgo, answeredInRound: big.NewInt(100),
+	}, maxAge, now)
+	if err != nil {
+		t.Fatalf("fresh round: unexpected error: %v", err)
+	}
+	if got.Cmp(price) != 0 {
+		t.Fatalf("fresh round: answer = %s, want %s", got, price)
+	}
+
+	// Each unhealthy round must be rejected; the answer must never leak through.
+	bad := []struct {
+		name  string
+		round chainlinkRound
+	}{
+		{"stale: updated 2h ago, max 1h", chainlinkRound{big.NewInt(100), price, twoHoursAgo, big.NewInt(100)}},
+		{"negative price", chainlinkRound{big.NewInt(100), big.NewInt(-1), tenMinAgo, big.NewInt(100)}},
+		{"zero price", chainlinkRound{big.NewInt(100), big.NewInt(0), tenMinAgo, big.NewInt(100)}},
+		{"round not complete (updatedAt 0)", chainlinkRound{big.NewInt(100), price, big.NewInt(0), big.NewInt(100)}},
+		{"carried-over answer (answeredInRound < roundId)", chainlinkRound{big.NewInt(100), price, tenMinAgo, big.NewInt(99)}},
+		{"nil answer", chainlinkRound{big.NewInt(100), nil, tenMinAgo, big.NewInt(100)}},
+	}
+	for _, tt := range bad {
+		t.Run(tt.name, func(t *testing.T) {
+			if got, err := validateChainlinkRound(tt.round, maxAge, now); err == nil {
+				t.Fatalf("expected error, got answer %s", got)
+			}
+		})
+	}
+}
+
+func TestChainlinkMaxAge(t *testing.T) {
+	if d := chainlinkMaxAge(nil); d != chainlinkDefaultMaxAge {
+		t.Fatalf("no arg: got %s, want default %s", d, chainlinkDefaultMaxAge)
+	}
+	if d := chainlinkMaxAge([]int{0}); d != chainlinkDefaultMaxAge {
+		t.Fatalf("zero arg: got %s, want default %s", d, chainlinkDefaultMaxAge)
+	}
+	if d := chainlinkMaxAge([]int{3600}); d != time.Hour {
+		t.Fatalf("explicit arg: got %s, want 1h", d)
+	}
+}
--- a/core/taskengine/macros/contract_test.go
+++ b/core/taskengine/macros/contract_test.go
@@ -1,6 +1,7 @@
 package macros
 
 import (
+	"fmt"
 	"math/big"
 	"testing"
 
@@ -45,7 +46,12 @@ func TestExpression(t *testing.T) {
 	rpc := testutil.GetTestRPCURL()
 	SetRpc(rpc)
 
-	p, e := CompileExpression(`priceChainlink("0x694AA1769357215DE4FAC081bf1f309aDC325306")`)
+	// Pass the max-age override so these live-feed assertions don't flake when the
+	// Sepolia feed is stale (the case guarded against in the staleness logic). This
+	// also exercises the optional maxAgeSeconds argument through the expr path.
+	price := fmt.Sprintf(`priceChainlink("0x694AA1769357215DE4FAC081bf1f309aDC325306", %d)`, testNoStaleBound)
+
+	p, e := CompileExpression(price)
 	if e != nil {
 		t.Errorf("Compile expression error: %v", e)
 	}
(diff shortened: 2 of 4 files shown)


chrisli30:
Thanks @Antrikshgwal, let us take a look at this PR!

chrisli30:
Thanks for the writeup in both the issue and the PR, @Antrikshgwal

The direction here is right, and pulling validateChainlinkRound out as a pure, RPC-free function is exactly how I'd want this tested. The six rejection cases plus the fresh-round pass cover the logic well.

A few things before we merge:

**Behavior change — I'm good with the direction**, just want it explicit. Re-pointing `priceChainlink/chainlinkPrice` at `latestRoundData` and panicking on a stale round means any existing task using those on a feed that's been stale >24h will now fail loudly instead of returning a price. That's the correct semantics and what #661 asked for, so let's do it — but please call it out in the PR description as a user-visible change so we can note it for SDK consumers.

Please also **fix contract_test.go**. `TestExpression` still calls `priceChainlink("0x694A…")` against the live Sepolia feed, which now routes through the new 24h default — that test will flake/panic whenever the testnet feed is stale, which is exactly the case you (correctly) guarded against in exp_test.go. Mirror the same large-maxAge there, or restructure so the live-feed tests don't gate on testnet freshness.

Minor: `100*365*24*3600` overflows a 32-bit int as an untyped constant — fine on our amd64 builds, but worth an int64 cast or a smaller sentinel. And if it's easy, a case exercising the expr path with the override `(priceChainlink(feed, 3600))` would lock in that the optional arg flows through the variadic correctly.

Antrikshgwal:
> Thanks for the writeup in both the issue and the PR, @Antrikshgwal
> 
> The direction here is right, and pulling validateChainlinkRound out as a pure, RPC-free function is exactly how I'd want this tested. The six rejection cases plus the fresh-round pass cover the logic well.
> 
> A few things before we merge:
> 
> **Behavior change — I'm good with the direction**, just want it explicit. Re-pointing `priceChainlink/chainlinkPrice` at `latestRoundData` and panicking on a stale round means any existing task using those on a feed that's been stale >24h will now fail loudly instead of returning a price. That's the correct semantics and what #661 asked for, so let's do it — but please call it out in the PR description as a user-visible change so we can note it for SDK consumers.
> 
> Please also **fix contract_test.go**. `TestExpression` still calls `priceChainlink("0x694A…")` against the live Sepolia feed, which now routes through the new 24h default — that test will flake/panic whenever the testnet feed is stale, which is exactly the case you (correctly) guarded against in exp_test.go. Mirror the same large-maxAge there, or restructure so the live-feed tests don't gate on testnet freshness.
> 
> Minor: `100*365*24*3600` overflows a 32-bit int as an untyped constant — fine on our amd64 builds, but worth an int64 cast or a smaller sentinel. And if it's easy, a case exercising the expr path with the override `(priceChainlink(feed, 3600))` would lock in that the optional arg flows through the variadic correctly.

Thanks @chrisli30. Addressed in the latest push:

- contract_test.go — TestExpression now passes the large max-age override to all three priceChainlink calls, so the live-feed assertions no longer gate on Sepolia freshness. This doubles as the override-through-expr coverage you suggested (priceChainlink(feed, N) runs through CompileExpression/RunExpressionQuery).

- int32 overflow — replaced with a shared testNoStaleBound (~50y) kept under MaxInt32.
- Behavior change — called out explicitly in the PR description for SDK consumers.

Antrikshgwal:
Unrelated, and no rush: when you have a spare minute, I'd value your read on the Calibur PoC scope over in #658 before I start building, so I aim the six-tx demo at what's most useful to you.

will-dz:
LGTM, this closes #661 cleanly. 

Validation is extracted and unit-tested without RPC, silent 0 / stale-price failure modes are gone, and the priceChainlink/chainlinkPrice re-point + panic-on-invalid behavior is the right tradeoff (and called out for SDK consumers).

Good to merge. I’ll add small pure-unit edge-case expansions in a follow-up PR. Thanks, @Antrikshgwal !

PR #699, opened 2026-08-01
fix: emit only the EIP-712 domain fields the type declares

chrisli30:
The prepare response's typed data carried `name`, `version` and `salt` as empty strings, although its `EIP712Domain` declares only `chainId` and `verifyingContract`:

```json
"types":  { "EIP712Domain": [ {"name":"chainId",...}, {"name":"verifyingContract",...} ] }
"domain": { "chainId":"0xaa36a7", "name":"", "salt":"", "verifyingContract":"0x209e…", "version":"" }
```

go-ethereum's `TypedDataDomain` has no `omitempty`, so marshalling it emits every field whether the domain uses one or not.

## Why nothing caught this

**The digest was never wrong.** EIP-712 verifiers read the domain through its *type*, so undeclared keys are ignored and the hash is unaffected. The existing round-trip test passes because it re-parses the payload with the same library that produced it — go-ethereum agrees with itself.

Strict clients do not. ethers v6 validates the domain object and rejects `salt: ""`:

```
TypeError: invalid BytesLike value (argument="value", value="", code=INVALID_ARGUMENT)
```

That fails a grant **before the owner is ever asked to sign** — the wallet never opens. Found by running the ava-sdk-js grant flow against a live local gateway, not by any test in this repo.

## The fix

Filter the serialized domain to the fields `EIP712Domain` declares. Keyed off the declared type rather than a hardcoded deny-list, so a domain that legitimately carries a `name` or `version` later keeps it without another fix here.

The builder (`userop.DeferredActionTypedData`) was already correct — it only ever set `ChainId` and `VerifyingContract`. The bug lived entirely in serialization, which is why the regression test asserts on the **HTTP response** rather than on the builder. I verified it fails without the fix:

```
domain carries "name", which EIP712Domain does not declare; strict clients reject it
```

## Verification

With this branch built and running locally, the full SDK grant flow passes end to end — prepare → `eth_signTypedData_v4` → submit → signature verifies → stored `pending` → revoked. Before it, that test died at the signing step.

Worth merging ahead of the SDK work: anything building against staging still has the bug.

### DIFF
--- a/aggregator/rest/handlers_policies.go
+++ b/aggregator/rest/handlers_policies.go
@@ -8,9 +8,8 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/labstack/echo/v4"
-
 	"github.com/ethereum/go-ethereum/signer/core/apitypes"
+	"github.com/labstack/echo/v4"
 
 	"github.com/AvaProtocol/EigenLayer-AVS/aggregator/rest/generated"
 	restmw "github.com/AvaProtocol/EigenLayer-AVS/aggregator/rest/middleware"
--- a/aggregator/rest/handlers_policies.go
+++ b/aggregator/rest/handlers_policies.go
@@ -10,6 +10,8 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/labstack/echo/v4"
 
+	"github.com/ethereum/go-ethereum/signer/core/apitypes"
+
 	"github.com/AvaProtocol/EigenLayer-AVS/aggregator/rest/generated"
 	restmw "github.com/AvaProtocol/EigenLayer-AVS/aggregator/rest/middleware"
 	"github.com/AvaProtocol/EigenLayer-AVS/core/taskengine"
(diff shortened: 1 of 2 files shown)


chrisli30:
## Audit — approve the change, rebase the branch first

**Verdict:** the fix is correct, minimal, and worth landing. But the branch is stale, which makes the PR unreviewable as shown and risky to merge as-is. Rebase onto current `staging`, then it's a clean one-commit PR.

### The fix is right ✅

- **Correctness of the approach.** Filtering the serialized domain to the fields `EIP712Domain` *declares* (rather than a hardcoded `name`/`version`/`salt` deny-list) is the right call — a domain that legitimately adds `name`/`version` later keeps it with no further change here.
- **It cannot change what the client signs.** The domain separator is hashed from the *declared* type fields only (`chainId`, `verifyingContract`), so dropping the empty extras leaves the signed digest identical. Confirmed: `TestDeferredActionDigestMatchesTypedData` (the wallet-parity guard) still passes. So there's zero risk of the client computing a different hash than the gateway — the one property that would have made this dangerous.
- **The regression test is placed correctly.** It asserts on the **HTTP response** a strict client actually receives, not on the builder (which was already correct). That's exactly where the bug lived, and it's the assertion the old same-library round-trip test structurally couldn't make. Verified it fails without the fix and passes with it; PR CI is green (18/18).
- Good diagnosis in the description — found by running the real SDK flow against a live gateway, which is the only way this surfaces (go-ethereum agreeing with itself hides it).

### 🔴 Blocking (hygiene): the branch is stale — rebase before merge

The merge-base of this branch with `staging` is `56315d4f` (#692), i.e. **before #694/#695/#697 landed**. That's why the PR shows **93 files / +9440** — it's re-displaying already-merged content. The **true net change is 2 files / +60**:

```
aggregator/rest/handlers_policies.go      | 27 +
aggregator/rest/handlers_policies_test.go | 33 +
```

Merging this branch as-is would re-introduce #694, #695, and a duplicate of #697 (`2b70e51d`) as separate commits into `staging`'s graph — recreating exactly the squash-divergence that was just cleaned up with `sync-main`, and risking conflicts against the squashed versions already on `staging`/`main`.

**Fix:** rebase onto current staging so the PR is just the one fix commit —

```
git rebase origin/staging   # drops #694/#695/#697 (already upstream), keeps 44ad6dbe
git push --force-with-lease
```

After that the diff is the 2 files above, history is clean, and the merge is trivial.

### Minor

- The `aggregator/rest` package intermittently `[setup failed]` under parallel local runs (BadgerDB temp-dirs / shared setup) — **pre-existing, not from this change** (CI is green; the fix touches no test setup or globals). Worth a separate test-isolation pass someday.
- `filterToDeclaredFields` is tidy and self-contained; no notes.

*Reviewed with Claude Code.*

PR #716, opened 2026-08-07
fix(taskengine): session grants replace instead of stacking

### LINKED ISSUE #715: Session policies: atomic supersede / replace-on-submit for ≤1 usable grant per capability
## Summary

Studio enforces **≤1 usable session policy per smart wallet + chain + capability** (e.g. Uniswap Auto grants) at the product layer. The gateway already **refuses execute** when more than one usable policy exists on a runner:

```text
wallet %s has more than one usable session policy (%s, %s); revoke one before executing
```

(`core/taskengine/session_policy.go`)

What the gateway **does not** provide is an **atomic** “install new grant and revoke the others” operation. Clients today must:

1. `policies.submit` (new grant becomes usable)
2. `policies.list`
3. `policies.revoke` for each other usable policy (N separate calls)

That multi-step flow is **not atomic**. Concurrent grant sessions (two tabs, two devices, Studio + future CLI/SDK client) can race and leave the runner with:

- **>1 usable** policies (execute still fails), or  
- **0 usable** policies (each side revokes the other’s keep id)

Studio mitigates with fail-closed post-submit verify (do not show “Authorized” unless a singleton remains) + Manage UX, but **cannot guarantee** the invariant under concurrent writers. We want that guarantee at the gateway.

**Product rule (locked for Ava Auto / Uniswap v1):**  
≤1 **usable** session policy per `(runner, chainId, capability)`. “Update permission” = replace, not stack.

Related Studio work: PR [AvaProtocol/studio#1458](https://github.com/AvaProtocol/studio/pull/1458), cardinality note `docs/changes/20260806-agent-permission-cardinality.md` (Studio repo). Live failure: dual ids on one Sepolia runner blocked Auto after a second successful grant sign.

---

## Problem

| Layer | Behavior today |
| --- | --- |
| **Execute** | Fail if `count(usable policies on runner) > 1` — safety backstop only |
| **Prepare / submit** | Allowed to stack a second (third, …) usable policy |
| **Revoke** | Per-id; no transactional multi-revoke or “revoke all others” |
| **Studio** | Best-effort supersede after submit + re-list verify; race remains |

**Race (two concurrent submits on the same runner):**

1. Client A submits policy `P_a` (keep = `P_a`), lists, starts revoking others.  
2. Client B submits policy `P_b` (keep = `P_b`), lists, starts revoking others.  
3. A revokes `P_b`, B revokes `P_a` → **zero** usable policies, or revoke failures leave **two**.

Neither client can fix this with extra retries alone without a single writer lock or server-side atomic replace.

Non-Studio clients (scripts, future partners) that call prepare/submit without supersede will **always** recreate dual-policy execute failures.

---

## Requested capability

Provide a **gateway-enforced** path so that after a successful grant for a given capability class, the runner ends with **exactly one** usable policy for that class (the kept one), even under concurrent clients.

#### Option A — `submit` with `replace` / supersede (recommended for v1 product)

Extend prepare and/or submit so the client can request replacement semantics, e.g.:

- `replaceExisting: true` (or default-on for a capability), and/or  
- `supersedePolicyIds: [...]` validated server-side, and/or  
- capability-scoped: “after this submit succeeds, revoke every other **usable** policy on this `(owner, runner, chainId)` that matches the same capability fingerprint”

**Atomicity requirement:** from the client’s POV, either:

- the new policy is stored **and** all superseded siblings are non-usable in one logical operation (single storage transaction / single writer critical section), **or**  
- the whole submit fails and no new usable policy is left without cleanup.

Revoking “others” must use the **same runner** that received the submit (not a second “primary wallet” resolution).

#### Option B — dedicated `policies.supersede` / `policies.replace`

```text
POST …/policies/supersede
{ runner, chainId, keepPolicyId, match: "uniswap_v3" | allowlistFingerprint | capabilityId }
```

Semantically: revoke all usable matching policies except `keepPolicyId`, under a lock held for the duration of list+revokes. Still requires Studio to call this immediately after submit unless combined with A.

#### Option C — reject second prepare/submit unless replace

Fail prepare/submit with a typed error when another usable same-capability policy already exists, unless `replace=true`. Forces explicit supersede path and protects non-Studio clients.

**Not requested for v1:** execute-time policy selection among multiple concurrent grants (multi-agent / multi-budget). That product choice would **remove** the current “fail if >1” invariant and needs a different design.

---

## Atomicity / concurrency expectations

1. **Critical section** keyed by `(owner, runner, chainId)` (or finer if capability is on the wire).  
2. Under that lock: list usable → apply keep/supersede rules → persist new grant and revoked statuses → unlock.  
3. Concurrent second submit either waits and then supersedes the first’s keep, or fails with a typed conflict so the client re-runs prepare (fresh entity) + submit with replace.  
4. No window where execute can see two usable policies after a successful replace response.  
5. If on-chain install/cleanup for revoked policies is async, **off-chain usability** (gateway `Usable()` / pending|active vs revoked) must still flip atomically in storage so execute resolution is consistent.

---

## API / error ergonomics (nice to have)

| Item | Why |
| --- | --- |
| Typed error code e.g. `SESSION_POLICY_AMBIGUOUS` / `MULTIPLE_USABLE_POLICIES` | Studio currently string-matches gateway prose; typed code is stable across clients |
| Optional list filters `usableOnly`, `capability=` | Cleaner Manage / supersede without client-side fingerprinting |
| Return `revokedPolicyIds[]` on replace/submit | Client logging + UX without a second list |

Capability is not yet a first-class field on `SessionPolicy` in the gateway; Studio fingerprints Uniswap via `exactInputSingle` (`0x04e45aaf`) on the allowlist. Prefer either:

- storing `capabilityId` (or grant class) on the policy at prepare, or  
- documenting a stable server-side fingerprint shared with Studio/SDK.

---

## Acceptance criteria

- [ ] A single successful “replace grant” API path leaves **exactly one** usable policy for that capability on the runner, even when two clients submit concurrently.  
- [ ] Execute continues to refuse `>1` usable as a backstop (defense in depth).  
- [ ] Documented contract for Studio/SDK: when to set replace, what errors mean, what to show the user.  
- [ ] Unit/integration tests covering: sequential replace; concurrent double-submit; revoke failure surfaces as failed replace (no silent dual-usable success).  
- [ ] No requirement for the client to issue N independent revokes to achieve the singleton (except optional best-effort cleanup for legacy dual rows already in the wild).

---

## Out of scope / non-goals

- Multi-concurrent usable grants with execute-time `policyId` selection (later product).  
- Changing ERC-4337 install mechanics beyond what is needed for off-chain usability atomicity.  
- Studio-only UX copy (already shipping in Studio).

---

## Context for implementers

| Resource | Note |
| --- | --- |
| Dual-policy execute guard | `core/taskengine/session_policy.go` (~“more than one usable session policy”) |
| Submit / revoke | `core/taskengine/engine_session_policy.go` (`SubmitSessionPolicy`, `RevokeSessionPolicyByID`) |
| REST | `aggregator/rest/handlers_policies.go` |
| Model | `model/session_policy.go` (`Usable()` = pending \| active, not revoked, not expired) |
| Studio client mitigation | Supersede-after-submit + re-list fail-closed (`submitPolicyAction`); still races without gateway lock |

---

## Priority rationale

**P2:** Execute safety already prevents fund movement under dual policies. Impact is **reliability and UX** (false “Authorized”, Manage friction, stuck Auto after concurrent or repeated grants) and **correctness for any non-Studio writer**. Required for a trustworthy multi-client permission model; Studio alone cannot close the race.

chrisli30:
Closes #715.

## Summary

A smart wallet may carry only one usable session grant — the send path refuses to execute when it finds more than one. Nothing enforced that at **write** time, so a second grant did not extend the user's permission, it bricked the wallet until someone revoked one by hand. That is what blocked Auto on Sepolia after a second successful grant sign.

Submit now revokes every other usable grant on the runner, and reports what it replaced in `supersededPolicyIds`.

## The issue's framing was off, and it changed the design

**This is mostly not a concurrency bug.** Two grants in plain sequence were enough: the first takes entity 1, then `NextSessionEntityID` returns 2, the owner signs for entity 2, and submit's entity re-check passes because 2 *is* the next free one. The pre-existing `ErrSessionEntityTaken` guard protects entity uniqueness, not policy singleton-ness, so any two grants separated by a completed submit sail through it.

**Capability scoping came out of v1.** The issue asked for ≤1 usable grant per `(runner, chainId, capability)`, but `ActiveSessionPolicyForWallet` already enforces something stricter — ≤1 per runner, whatever the capability. Capability-scoped supersede would deliberately permit a pair that execute then rejects. Scoping to the runner instead removes the entire prerequisite: no `capabilityId` field, no server-side fingerprint, no mirroring Studio's `exactInputSingle` heuristic in the gateway.

**None of Options A/B/C as written.** A (`replaceExisting: true`) and C (reject-unless-replace) make correct behavior opt-in — a flag whose default value is "break" — and leave non-Studio writers that never learned the flag still broken. Replace is therefore unconditional, which fixes those clients with no change on their side and needs no request-shape change. B (a dedicated supersede endpoint) only existed to clean up legacy dual rows, which a plain re-grant now does.

Full reasoning: [design note on #715](https://github.com/AvaProtocol/EigenLayer-AVS/issues/715#issuecomment-5210598485).

## How it works

- `SubmitSessionPolicy` holds a write lock on `(chainID, owner, runner)` across the entity re-check, the store, and the supersede. `RevokeSessionPolicyByID` takes the same lock; the resolver and the contract-write preflight take it **shared**, so an executing workflow never observes the window between the two writes.
- An in-process sharded `RWMutex` suffices, and that is a property of the deployment rather than an assumption: Badger is embedded, opened once per aggregator, and holds an exclusive directory lock — a second writer process cannot exist.
- Because it supersedes *all* other usable grants, the operation is **self-healing**: wallets already stuck with dual rows are repaired by the next grant. No migration, no cleanup endpoint.

### Two details that bite

- **Superseding revokes, never deletes.** `RevokeSessionGrant` deletes pending policies, but `NextSessionEntityID` counts stored records — so a deleted record frees its entity. Superseding a pending grant whose install was in flight would hand that entity to the next grant, whose `installValidation` would land on an entity the account already has.
- **Store-then-supersede, not the reverse.** A crash between the two leaves two usable grants, which fails closed and self-heals. Superseding first could leave *zero* — working authority destroyed for a replacement that never landed.

## Client-visible changes

| Change | Kind |
| --- | --- |
| Submit revokes the runner's other usable grants | Behavior — no opt-in, no opt-out |
| `supersededPolicyIds` on the 201 (always present, empty on a first grant) | Additive |
| `SESSION_POLICY_AMBIGUOUS:` prefix on the >1-usable refusal | Additive; Studio currently string-matches the prose |
| `POLICIES_SUPERSEDE_FAILED` (500) — grant stored, previous not revoked; do **not** retry the submit, the entity is spent | New error |

`SubmitPolicyResponse` is `allOf: [SessionPolicy, {supersededPolicyIds}]`, a wire-compatible superset — the existing test that decodes the 201 as `SessionPolicy` still passes unchanged.

## Off-chain only

Superseding marks a grant unusable **by this gateway**. The validation entity and its ERC-20 spend cap stay installed until the owner signs `uninstallValidation`, so "update permission" does not reduce what the account *could* authorize on chain. Closing that half means batching the uninstall into the deferred action — real work on payload shape and signature semantics, deliberately out of scope, needs its own issue.

## Test plan

`core/taskengine/session_policy_supersede_test.go`, all under `-race`:

- [x] a second grant replaces the first and reports it (the sequential regression)
- [x] a superseded grant is retained as `revoked` and its entity is never recycled
- [x] replacement does not touch the owner's other wallets
- [x] a wallet seeded with two usable grants is repaired by granting again
- [x] the ambiguity refusal carries `SESSION_POLICY_AMBIGUOUS`
- [x] six concurrent submits: exactly one accepted, one usable grant remains
- [x] 25 sequential re-grants against a spinning resolver never expose the store-then-supersede window — **verified to fail 3/3 without the resolver's read lock**
- [x] a supersede failure surfaces as an error, not a 201

REST: HTTP round trip for replace, plus a drift guard pinning `SubmitPolicyResponse` against `SessionPolicy` so the hand-written converter for the flattened `allOf` cannot silently drop a field — **verified to fail when a field is removed**.

- [x] `make storage-check` — no key or model changes
- [ ] CI green
- [ ] SDK consumes `supersededPolicyIds`; Studio drops its supersede-after-submit mitigation

### DIFF
--- a/aggregator/rest/generated/server.gen.go
+++ b/aggregator/rest/generated/server.gen.go
@@ -1990,7 +1990,7 @@ type SubmitWalletPolicyResponseObject interface {
 	VisitSubmitWalletPolicyResponse(w http.ResponseWriter) error
 }
 
-type SubmitWalletPolicy201JSONResponse SessionPolicy
+type SubmitWalletPolicy201JSONResponse SubmitPolicyResponse
 
 func (response SubmitWalletPolicy201JSONResponse) VisitSubmitWalletPolicyResponse(w http.ResponseWriter) error {
 	w.Header().Set("Content-Type", "application/json")
--- a/aggregator/rest/generated/types.gen.go
+++ b/aggregator/rest/generated/types.gen.go
@@ -1486,7 +1486,7 @@ type SubmitPolicyResponse struct {
 	//
 	// Empty on a first grant. Non-empty means the user's earlier
 	// permission is gone, which is worth reflecting in the UI.
-	SupersededPolicyIds *[]Ulid `json:"supersededPolicyIds,omitempty"`
+	SupersededPolicyIds []Ulid `json:"supersededPolicyIds"`
 
 	// ValidUntil Unix milliseconds.
 	ValidUntil int64 `json:"validUntil"`
--- a/aggregator/rest/generated/types.gen.go
+++ b/aggregator/rest/generated/types.gen.go
@@ -202,8 +202,8 @@ const (
 
 // Defines values for RevokePolicyResponseStatus.
 const (
-	Deleted RevokePolicyResponseStatus = "deleted"
-	Revoked RevokePolicyResponseStatus = "revoked"
+	RevokePolicyResponseStatusDeleted RevokePolicyResponseStatus = "deleted"
+	RevokePolicyResponseStatusRevoked RevokePolicyResponseStatus = "revoked"
 )
 
 // Defines values for SecretScope.
@@ -224,6 +224,13 @@ const (
 const (
 	Approve SignalExecutionRequestDecision = "approve"
 	Reject  SignalExecutionRequestDecision = "reject"
+)
+
+// Defines values for SubmitPolicyResponseStatus.
+const (
+	SubmitPolicyResponseStatusActive  SubmitPolicyResponseStatus = "active"
+	SubmitPolicyResponseStatusPending SubmitPolicyResponseStatus = "pending"
+	SubmitPolicyResponseStatusRevoked SubmitPolicyResponseStatus = "revoked"
 )
 
 // Defines values for TokenMetadataResponseSource.
@@ -1438,6 +1445,56 @@ type SubmitPolicyRequest struct {
 	// ValidUntil The ABSOLUTE expiry from prepare. It is baked into the signed calldata; recomputing it would change the digest.
 	ValidUntil int64 `json:"validUntil"`
 }
+
+// SubmitPolicyResponse defines model for SubmitPolicyResponse.
+type SubmitPolicyResponse struct {
+	AgentLabel     string           `json:"agentLabel"`
+	AllowedActions *[]AllowedAction `json:"allowedActions,omitempty"`
+
+	// ChainId Numeric chain ID (e.g. 11155111 for Sepolia, 8453 for Base). On
+	// chain-aware trigger/node configs this is required and must be a
+	// configured chain; on query/filter params it is optional.
+	ChainId ChainId `json:"chainId"`
+
+	// CreatedAt Unix milliseconds.
+	CreatedAt int64 `json:"createdAt"`
+	EntityId  int64 `json:"entityId"`
+
+	// Erc20SpendCap Cumulative ERC-20 spend cap, enforced on-chain at execution. The
+	// token must appear as an `allowedActions` target.
+	Erc20SpendCap *Erc20SpendCap `json:"erc20SpendCap,omitempty"`
+
+	// Id ULID identifier (26-char Crockford base32).
+	Id            Ulid    `json:"id"`
+	Justification *string `json:"justification,omitempty"`
+
+	// Runner Lowercase or checksummed hex EOA / contract address.
+	Runner EthereumAddress `json:"runner"`
+
+	// SessionSigner Lowercase or checksummed hex EOA / contract address.
+	SessionSigner EthereumAddress `json:"sessionSigner"`
+
+	// Status pending = signed and stored, install not yet on-chain (revocable
+	// for free). active = install applied. revoked = grants nothing.
+	Status SubmitPolicyResponseStatus `json:"status"`
+
+	// SupersededPolicyIds Grants revoked to keep this runner's authority a singleton —
+	// the previous grants this submit replaced. Their `status` now
+	// reads `revoked`; the array distinguishes a replacement the
+	// gateway performed from one the user asked for via
+	// `DELETE .../policies/{policyId}`.
+	//
+	// Empty on a first grant. Non-empty means the user's earlier
+	// permission is gone, which is worth reflecting in the UI.
+	SupersededPolicyIds *[]Ulid `json:"supersededPolicyIds,omitempty"`
+
+	// ValidUntil Unix milliseconds.
+	ValidUntil int64 `json:"validUntil"`
+}
+
+// SubmitPolicyResponseStatus pending = signed and stored, install not yet on-chain (revocable
+// for free). active = install applied. revoked = grants nothing.
+type SubmitPolicyResponseStatus string
 
 // Timestamp RFC 3339 timestamp.
 type Timestamp = time.Time
--- a/aggregator/rest/handlers_policies.go
+++ b/aggregator/rest/handlers_policies.go
@@ -305,14 +305,14 @@ func submitPolicyToAPI(p *model.SessionPolicy, superseded []string) generated.Su
 		ValidUntil:     base.ValidUntil,
 		CreatedAt:      base.CreatedAt,
 	}
-	// Always present, empty on a first grant: a client checking "did this
-	// replace anything" should read an empty array, not have to treat a
-	// missing key as the same thing.
+	// Required in the schema and always emitted, empty on a first grant: a
+	// client asking "did this replace anything" reads an empty array rather
+	// than having to treat a missing key as the same thing.
 	ids := make([]generated.Ulid, 0, len(superseded))
 	for _, id := range superseded {
 		ids = append(ids, generated.Ulid(id))
 	}
-	out.SupersededPolicyIds = &ids
+	out.SupersededPolicyIds = ids
 	return out
 }
 
--- a/aggregator/rest/handlers_policies.go
+++ b/aggregator/rest/handlers_policies.go
@@ -136,7 +136,7 @@ func (s *Server) SubmitWalletPolicy(ctx echo.Context, address generated.Ethereum
 		return badRequest("POLICIES_BAD_PERMISSIONS", "Invalid permissions", err.Error())
 	}
 
-	policy, err := s.engine.SubmitSessionPolicy(user, taskengine.SessionPolicyInput{
+	policy, superseded, err := s.engine.SubmitSessionPolicy(user, taskengine.SessionPolicyInput{
 		Wallet:        wallet,
 		ChainID:       int64(req.ChainId),
 		AgentLabel:    req.AgentLabel,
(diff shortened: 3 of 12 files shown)


chrisli30:
CI: the two red jobs — `Unit Test (core/taskengine)` and `Unit Test (pkg/erc4337/preset)` — are **pre-existing on staging, not from this PR**.

#714 merged into staging before this branch was cut and contains none of this code. Its run fails the identical tests:

| Test | Job | Error | Fails on #714 |
| --- | --- | --- | --- |
| `TestUserOpETHWithdrawal_Sepolia` | core/taskengine | `estimating gas: eth_estimateUserOperationGas: AA23 reverted` | yes |
| `TestSendUserOp`, `TestUserOpExecutionSuccessWithPaymaster` | pkg/erc4337/preset | `AA31 paymaster deposit too low` | yes |

Both are live-Sepolia tests hitting real chain state, and this PR touches neither package's execution path — `pkg/erc4337/preset` is untouched entirely.

`AA31 paymaster deposit too low` is straightforwardly the Sepolia VerifyingPaymaster running dry and wants a top-up. The `AA23` in the withdrawal test is on the shared salt-0 runner `0x209eb31c199bEB4c386eF83CF442DE1a00667a1F`, whose entity 1 is already installed on chain ("not replaying the install") — worth a separate look at that wallet's on-chain grant state.

Everything this PR adds is green: all new supersede tests pass locally under `-race`, and `Unit Test (aggregator)` — which covers the REST changes here — passes in CI.

Follow-up filed as #717 for the on-chain half (batching `uninstallValidation` into the deferred action).

chrisli30:
Worked through the Copilot review — all three findings were real, fixed in c4d17764.

**`MarkSessionGrantApplied*` not taking the runner lock.** The important one. Re-reading before storing is not sufficient on its own: the confirmation reads `pending`, a submit supersedes the grant, and the confirmation then stores `active` over it — resurrecting a replaced grant and leaving the runner with two usable ones, which is exactly the state this PR exists to prevent. And because this PR makes replacement routine rather than rare, it moves that window from theoretical to reachable. `MarkSessionGrantAppliedByID` now takes the runner's write lock across the re-read and the write; it needed the runner threaded through, which the resolver now captures alongside the owner. Safe to take there because the resolver has released its read lock by the time `OnApplied` fires — nothing on that path holds it.

On testing it: I first wrote a sequential test, checked it, and it passed **without** the lock — it only exercised the pre-existing re-read guard. A concurrent version didn't catch it either: the confirmation's read-then-write is microseconds while the submit racing it spends far longer in signature recovery, so the interleaving effectively never happens on demand. Rather than ship a test that proves nothing, the test now asserts the property that actually changed — hold the runner's lock and the confirmation must wait for it. That fails 2/2 with the lock removed and passes with it. The genuine interleaving remains uncovered, and I would rather say so than imply otherwise.

**`allocation-free` comment was false.** Right — `FormatInt`, `Hex()` and `ToLower` all allocate. Rather than soften the comment I made it true: the shard hash now runs over the raw `common.Address` bytes, which also removes the checksum-casing question, since an address is already canonical 20 bytes and the hex round-trip was pure ceremony. Added a test pinning the property sharding depends on — the same runner must always reach the same mutex, since two writers on different locks would not exclude each other at all.

**`supersededPolicyIds` should be required.** Agreed. The server always emits it, empty on a first grant, so leaving it optional only cost clients a pointer and a nil check that could never be meaningful. Marked required, regenerated, and the handler and tests now use the value type.

`make storage-check` still clean; offline suites green under `-race`. CI running.

Issue #719, opened 2026-08-07
Live-chain test fixtures: per-test salt + verify-or-restore session permissions (TestUserOpETHWithdrawal_Sepolia AA23)


chrisli30:
## Summary

`TestUserOpETHWithdrawal_Sepolia` fails CI with `AA23 reverted` during `eth_estimateUserOperationGas`. Root cause confirmed on chain: **every live grant test shares one Sepolia wallet at salt 0, and `grantControllerAuthority` adopts whatever validation entity it finds there without checking that the entity's on-chain permissions match the grant it just modelled off-chain.**

Fix needs two things: per-test wallet isolation, and a fixture helper that verifies (or restores) the expected permission state instead of assuming it.

## Evidence

The shared runner is `0x209eb31c199bEB4c386eF83CF442DE1a00667a1F` (owner + salt 0). Reading it on Sepolia:

```
signers(entity, wallet) on SingleSignerValidationModule 0x…99DE0BF6:
  entity 1 → 0x82F2Dd9a552a69f2ceD7Ff2D05c43aB8430158FB   (the controller)
  entity 2 → 0x82F2Dd9a552a69f2ceD7Ff2D05c43aB8430158FB   (the controller)
  entity 3+ → unset
```

`getValidationData` for entity 1 (identical for entity 2):

| Field | Value |
| --- | --- |
| validationFlags | 5 |
| validationHooks | `AllowlistModule` 0x…3e826473 (entity 1), `TimeRangeModule` 0x…82B8e201 (entity 1) |
| **executionHooks** | **`AllowlistModule` 0x…3e826473 (entity 1), flags 0x04** |
| selectors | none (global validation) |

Both occupied entities are **production-shaped Studio grants** — allowlist + time range + an allowlist *execution* hook. Neither is a bare grant.

## Why that produces AA23

`grantControllerAuthority` (`session_grant_testhelper_test.go`) builds a **bare global grant with no hooks** (`Selectors: nil, Hooks: nil, AllowSelfAdministration: true`) against a fresh in-memory DB, so `NextSessionEntityID` always returns **1**. It then does:

```go
if installedOnChain(t, swCfg, wallet, policy.EntityID, controller) {
    MarkSessionGrantApplied(db, policy, "pre-existing")   // adopt whatever is there
}
```

[...]

Issue #718, opened 2026-08-07
Retire the v0.6 send path and the self-funded VerifyingPaymaster (code + tests)


chrisli30:
## Summary

Every chain runs Modular Account v2 on EntryPoint v0.7, and sponsorship comes from the Alchemy Gas Manager policy. The v0.6 send path and the self-funded `VerifyingPaymaster` are retired **in deployment** but still present **in code** — and their live tests fail CI on every PR.

Remove the path and its tests together.

## Why now

`Unit Test (pkg/erc4337/preset)` has been red on every PR (#714, #716, staging) with:

```text
error sending transaction to bundler: revert reason : AA31 paymaster deposit too low
```

That is the Sepolia `VerifyingPaymaster` EntryPoint deposit running dry. Topping it up would buy a green run for a path nothing in production uses; the deposit would drain again and the tests would go back to red.

The codebase already states the intended resolution, in `pkg/erc4337/preset/builder_test.go`:

> `SendUserOp` is the v0.6 path — v0.6 EntryPoint, `executeBatchWithValues` calldata … **When the v0.6 path is removed post-cutover, this test goes with it.**

Note the order: the path goes, and the test follows. Deleting the tests alone would leave a config-selectable production branch with no coverage at all — a silent trapdoor rather than a loud failure.

## What is still live

`SendUserOpAuto` (`pkg/erc4337/preset/sent_userop.go`) dispatches per chain and still carries the v0.6 branch:

```go
if smartWalletConfig.UsesModularAccountV2() {
    ... SendUserOpMAv2(...)
}
return SendUserOp(smartWalletConfig, owner, callData, paymasterReq, senderOverride, saltOverride, executionFeeWei, lgr)
```

Reachable from five call sites — `vm_runner_contract_write.go`, `vm_runner_eth_transfer.go`, `aggregator/rpc_server.go` (×2), `worker/server.go` — whenever a chain sets `account_provider: simple_account`. No deployed chain does, and `config/gateway.example.yaml` defaults to `modular_account_v2`, but nothing prevents it.

## Scope

Remove, in one PR:

- `SendUserOp`, `SendUserOpWithWsClient` (`pkg/erc4337/preset/builder.go`)
- The v0.6 branch of `SendUserOpAuto` / `SendUserOpAutoWithWsClient`, which collapses to the MA v2 call, plus `sentFromV06`
[...]

Issue #717, opened 2026-08-07
Session grants: batch uninstallValidation into the deferred action so revoke/replace reaches the chain


chrisli30:
## Summary

Revoking or replacing a session grant is **off-chain only**. The validation entity and its permission hooks — ERC-20 spend cap, time range — stay installed on the Modular Account v2 runner until the owner signs `uninstallValidation`, and nothing in the product ever asks them to.

This is the other half of #715 (gateway PR #716), which made submit replace the runner's previous grants. That fixed which grant the gateway *will use*. It did not change what the account *could authorize*.

## What this looks like in practice

After three re-grants on one wallet:

| | Off-chain (gateway) | On-chain (account) |
| --- | --- | --- |
| Usable grants | 1 | — |
| Installed validation entities | — | 3 |
| Live ERC-20 spend caps | 1 shown | 3, each with its own remaining allowance |

So "update permission" in Studio does not reduce the user's exposure, it adds to it. A user who lowers a cap from 500 to 100 USDC has 600 USDC of live on-chain authority, of which our gateway will voluntarily use 100. Every entity is signed by the same shared controller key by design (`controllerSessionSigner`), so a compromised controller reaches all of them, not just the current one.

The same gap already exists on the explicit revoke path: `RevokeSessionGrant` returns `onChainCleanupRequired = true` for an applied grant and deliberately retains the record so the cleanup payload can be rebuilt — and then nothing rebuilds it. `SessionPolicy.Grant.InstallCall` is stored for exactly this purpose (see the type comment in `model/session_policy.go`).

## Proposal

Make the grant's deferred action a **batch**: `uninstallValidation(prior entity)` followed by `installValidation(new entity)`, committed to by the owner's single existing signature. Replacement then means the same thing on both sides, with no extra prompt.

Sketch, against the current flow in `core/taskengine/session_grant.go`:

- `PrepareSessionGrant` resolves the runner's current usable-and-applied grant, and when there is one, packs uninstall+install instead of install alone. `Grant.InstallCall` on the prior policy is what the uninstall tuple is rebuilt from.
- The digest already commits to the whole call, so the owner signs the replacement atomically — no second signature, no second screen.
- `SubmitSessionGrant`'s re-check extends from "is my entity still the next free one" to "is the grant I am superseding still the one I prepared against". Same shape as the existing `ErrSessionEntityTaken` path: if it moved, prepare again.

## Things that need deciding, not just implementing

[...]

PR #720, opened 2026-08-07
fix(test): verify fixture permissions on chain instead of adopting any entity

chrisli30:
Refs #719 — does NOT close it: the isolation criterion (live grant tests not sharing a wallet) needs funded distinct salts, which this PR sets up but does not flip.

## Summary

Live grant fixtures asked the account *"does this validation entity name my signer"* and treated a yes as *"this entity is the grant I need"*. Those come apart exactly when it matters, and that is why `TestUserOpETHWithdrawal_Sepolia` has been failing CI with `AA23 reverted`.

Fixtures now declare the permission state they need, and the helper converges the runner on it.

## Root cause, read off the chain

Entity 1 on the shared runner `0x209eb31c199bEB4c386eF83CF442DE1a00667a1F`:

| | |
| --- | --- |
| signer | `0x82F2Dd9a…0158FB` — the controller, so the old check passed |
| validationHooks | `AllowlistModule` #1, `TimeRangeModule` #1 |
| **executionHooks** | **`AllowlistModule` #1** |
| selectors | none (global) |

Entity 2 is identical. Both are Studio-shaped grants; there was no bare entity on that wallet at all.

`grantControllerAuthority` builds a **bare hookless** grant against a fresh in-memory DB, so it always lands on entity 1, saw the controller, and adopted it. The operation it then builds is rejected twice over:

1. **Not `executeUserOp`-wrapped.** A bare grant stores `RequiresExecuteUserOp=false`, but the real entity carries an execution hook, and every operation under such a grant must be wrapped.
2. **Denied by the allowlist**, which was installed for a Uniswap/USDC grant and does not cover a plain ETH transfer to the owner EOA.

#714 documented this exact hazard and fixed it for the production-shaped helper only. The bare helper never got the same treatment, which is how it stayed broken.

## What changed

- `expectedPermissions` — the "expected permission object" a fixture declares: signer, validation hooks, execution hooks, global-vs-scoped. Hook sets compare as sets, since the account stores hooks prepend-on-add.
- `readValidationState` reads the whole state: the signer from the `SingleSignerValidationModule`, the hooks and selector scoping from the account's own `getValidationData`.
- `resetInitialPermissions` converges the runner: **reuse** an entity that matches exactly (mark the stored grant applied so the install is not replayed), **step over** one that does not, **claim** the first free one. It fails with an expected-vs-actual diff naming the modules rather than sending the reader to a block explorer.
- **Both** grant helpers run through it, so the bare and production-shaped fixtures cannot drift apart again.
- `installedOnChain` is deleted rather than left available — it is the bug, not an optimisation.

### Converge, not wipe

Wiping is not available. Uninstalling a stale entity needs each hook's `onUninstall` payload, and the `AllowlistModule`'s is the `(entityId, inputs)` tuple it was installed with — recoverable only from the stored install call, as `core/chainio/aa/ma_v2_uninstall.go` is explicit about. A fixture wallet's stale entities came from runs whose database is long gone, so that payload cannot be reconstructed, and uninstalling without it strands module state instead of clearing it.

This is why the issue's preferred option (b) is not what shipped.

Steady state on a wallet used by one fixture is a **single entity**: run one installs it, every later run matches and adopts it. Entities accumulate only when a fixture's expected shape changes — a code change, not a per-run cost.

## Salts: mechanism shipped, flip deferred

Each fixture now derives its runner through a named salt constant, and `requireFundedRunner` hard-fails with the runner address and the exact shortfall instead of skipping (a skipped live test reads as a passing suite that never exercised the path).

**The salts are all still 0.** A fresh salt is an undeployed, unfunded counterfactual wallet, so flipping one without funding it turns a passing live test into a failing one. Isolating a fixture is now: change one constant, run once, fund the address the preflight prints. Happy to do that in a follow-up once someone confirms the funding.

Verification is what fixes the red test; salt isolation is hardening on top.

## Test plan

- [x] `decodeValidationData` tested offline against the **verbatim `getValidationData` return captured from that Sepolia runner** — the comparison that decides all of this is checked against what the account really says, not a hand-built fixture.
- [x] The regression in one assertion: the signer matches (all the old check looked at) and `bareGlobalGrant` still refuses the entity.
- [x] The production-shaped fixture adopts a matching entity, so wallets do not grow an entity per CI run.
- [x] Diffs: free entity, foreign signer, hook order, hooks at another entity, selector scoping.
- [x] `go build ./...`, `go vet`, offline suites green under `-race`.
- [ ] CI green
- [ ] `TestUserOpETHWithdrawal_Sepolia` passes live — **not verifiable locally** (needs `TEST_PRIVATE_KEY`, `OWNER_EOA` and a funded Sepolia runner). Expected path: entities 1 and 2 mismatch and are stepped over, entity 3 is free and claimed, and the install rides the first operation with genuinely bare permissions.

Independent of #716 (session-grant replace) despite touching the same package.

### DIFF
--- a/core/taskengine/execute_sequential_contract_writes_test.go
+++ b/core/taskengine/execute_sequential_contract_writes_test.go
@@ -29,7 +29,7 @@ func TestExecuteTask_SequentialContractWrites_Sepolia(t *testing.T) {
 
 	ownerAddr, ok := testutil.MustGetTestOwnerAddress()
 	if !ok {
-		t.Skip("Owner EOA address not set, skipping real execution test")
+		t.Fatal("OWNER_EOA must be set: this test executes against that owner's smart wallet")
 	}
 	ownerAddress := *ownerAddr
 
@@ -61,9 +61,10 @@ func TestExecuteTask_SequentialContractWrites_Sepolia(t *testing.T) {
 	require.NoError(t, err, "Failed to get wallet balance")
 	t.Logf("   ETH Balance: %s wei", balance.String())
 
-	if balance.Cmp(big.NewInt(0)) == 0 {
-		t.Skip("Smart wallet has 0 ETH balance - fund it to run real execution test")
-	}
+	// The wallet reimburses the paymaster for gas, so it needs a non-zero
+	// balance. Hard-fail: a skipped live test reads as a passing suite that
+	// never exercised the path.
+	requireFundedRunner(t, cfg.SmartWallet, *smartWalletAddr, big.NewInt(1))
 
 	// Initialize test database and engine
 	db := testutil.TestMustDB()
--- a/core/taskengine/execute_sequential_contract_writes_test.go
+++ b/core/taskengine/execute_sequential_contract_writes_test.go
@@ -83,7 +83,7 @@ func TestExecuteTask_SequentialContractWrites_Sepolia(t *testing.T) {
 		Owner:   &ownerAddress,
 		Address: smartWalletAddr,
 		Factory: &cfg.SmartWallet.FactoryAddress,
-		Salt:    big.NewInt(0),
+		Salt:    big.NewInt(fixtureSaltSequentialWrites),
 	})
 	require.NoError(t, err, "Failed to register smart wallet")
 	t.Logf("   ✅ Smart wallet registered in database")
--- a/core/taskengine/execute_sequential_contract_writes_test.go
+++ b/core/taskengine/execute_sequential_contract_writes_test.go
@@ -52,7 +52,7 @@ func TestExecuteTask_SequentialContractWrites_Sepolia(t *testing.T) {
 	defer client.Close()
 
 	// Compute the salt:0 smart wallet address for this owner
-	smartWalletAddr, err := aa.GetSenderAddress(client, ownerAddress, big.NewInt(0))
+	smartWalletAddr, err := aa.GetSenderAddress(client, ownerAddress, big.NewInt(fixtureSaltSequentialWrites))
 	require.NoError(t, err, "Failed to derive smart wallet address")
 	t.Logf("   ✅ Smart Wallet (salt:0): %s", smartWalletAddr.Hex())
 
--- a/core/taskengine/session_fixture_permissions_decode_test.go
+++ b/core/taskengine/session_fixture_permissions_decode_test.go
@@ -36,15 +36,17 @@ func TestDecodeValidationDataFromSepolia(t *testing.T) {
 	require.NoError(t, err)
 
 	require.Equal(t, []hookRef{
-		{Module: aa.AllowlistModuleAddress(), Entity: 1},
-		{Module: aa.TimeRangeModuleAddress(), Entity: 1},
+		{Module: aa.AllowlistModuleAddress(), Entity: 1, Flags: aa.HookFlagValidation},
+		{Module: aa.TimeRangeModuleAddress(), Entity: 1, Flags: aa.HookFlagValidation},
 	}, state.ValidationHooks, "entity 1 carries the allowlist and time-range validation hooks")
 
 	require.Equal(t, []hookRef{
-		{Module: aa.AllowlistModuleAddress(), Entity: 1},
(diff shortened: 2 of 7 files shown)


chrisli30:
CI confirms the diagnosis. `Unit Test (core/taskengine)` is **green**, and `TestUserOpETHWithdrawal_Sepolia` passes for the first time — with the helper narrating exactly the predicted path:

```
fixture: entity 1 on 0x209eb31c… is not usable — validation hooks are
  [0x…82B8e201#1 0x…3e826473#1], want []; execution hooks are [0x…3e826473#1], want []; trying the next one
fixture: entity 2 on 0x209eb31c… is not usable — … ; trying the next one
fixture: entity 3 on 0x209eb31c… is free; the install rides the first operation
--- PASS: TestUserOpETHWithdrawal_Sepolia (24.43s)
```

Entities 1 and 2 carry the TimeRange + Allowlist hook set the old signer-only check adopted blindly; entity 3 was free and got a genuinely bare grant. This was the one claim in the PR description I could not verify locally, so consider that box ticked.

`Unit Test (pkg/erc4337/preset)` is still red on `AA31 paymaster deposit too low` — unrelated to this PR, unchanged from staging, and tracked in #718.

---

Second commit applies the hard-fail rule uniformly across the live fixtures, per review feedback: owner address, config load, RPC reachability and every balance check (native ETH, ERC-20, EntryPoint deposit) now fail instead of skipping, routed through `requireFundedRunner` / `requireFixtureBalance` so the message names the address and the shortfall.

Selectors are deliberately left as skips — `-short` means "not running the live tests" and `TEST_CHAIN` / chain-id checks mean "this test is not about the chain you pointed me at". Neither is a missing prerequisite.

One consequence worth knowing: `make test` does not pass `-short`, so a local run without `OWNER_EOA` / `TEST_PRIVATE_KEY` now fails on live tests where it used to skip. CI passes both `-short` and the credentials, so it is unaffected.

chrisli30:
Worked through the Copilot review — four findings were real, and one was a regression this PR introduced. Fixed in f5f0cee6.

**1. Adopting stateful hooked entities (the regression).** Copilot is right, and this was the serious one. A `HookConfig` names a module and an entity and says nothing about what that module was configured to permit — an allowlist installed for USDC-only presents an identical HookConfig to one installed for USDC+WETH. `grantControllerAuthorityWithERC20Approves` previously skipped *every* occupied entity and always installed fresh; my convergence logic relaxed that to "reuse on shape match", which would have reintroduced precisely the stale-allowlist AA23 that #714 fixed.

Now gated on `expectedPermissions.Reusable`: only `bareGlobalGrant` is reusable, because "no hooks" is a complete description with no module state hiding behind it. The hooked fixture steps over matching entities and installs fresh, keeping #714's behavior while gaining the readable diff. Note this also gives up the "steady state is one entity" property for the hooked fixture — correctness first; bounded growth is what `maxFixtureEntityScan` and #717 are for.

**2. `validationFlags` decoded and discarded.** Right — global authority is not inferable from an empty selector list. An entity with `UserOp` but not `Global` authorizes no global call and would have matched; one carrying `Signature` holds signing authority no fixture asked for. Flags are now stored and compared against `ValidationFlagUserOp|ValidationFlagGlobal`.

**3. Dropped HookConfig flag byte.** Right in principle — the same module registered post-only is not the authorization a pre-execution hook provides. `hookRef` now carries `Flags` and compares them.

**4. `entitySignerOnChain` folding errors into the zero address.** Right, and it was inconsistent with my own `getValidationData` handling. Zero means "free to install on", so a transient RPC failure could have installed a second validation over an existing one. It now returns the error and the caller hard-fails.

**5–9. Salt not threaded through deployment and `SmartWallet.Salt`.** Right, and it made this PR's central claim false: only address *derivation* used the constant, so flipping one would derive an address for salt N while deploying and recording salt 0, leaving the engine injecting `aa_salt=0` for a runner derived from another salt. All three sites per fixture now use the constant, so the one-line flip actually works.

**10. `Closes #719` with the isolation criterion unmet.** Fair. Changed to `Refs #719` so the issue stays open for the funded-salt work.

New tests cover each: reuse gating, validation-flag comparison (missing `Global`, unexpected `Signature`), and hook callback flags. Offline suites green under `-race`; short mode still skips live tests cleanly.

PR #721, opened 2026-08-07
chore: retire the v0.6 send path and the self-funded VerifyingPaymaster

### LINKED ISSUE #718: Retire the v0.6 send path and the self-funded VerifyingPaymaster (code + tests)

chrisli30:
Closes #718.

## Summary

Every chain runs Modular Account v2 on EntryPoint v0.7 and sponsors gas through the Alchemy Gas Manager policy. The v0.6 send path was retired in deployment but still present in code. **~3,200 lines removed, ~160 added.**

`SendUserOpAuto` still dispatched into it from five call sites — contract write, ETH transfer, `rpc_server` (×2), `worker/server` — whenever a chain set `account_provider: simple_account`. No deployed chain did, but nothing stopped one, and the branch targeted an EntryPoint the fleet no longer transacts against.

The codebase already recorded the intended order, in `builder_test.go`: *"When the v0.6 path is removed post-cutover, this test goes with it."* — the path first, the test following. Deleting the tests alone would have left a config-selectable production branch with no coverage: a silent trapdoor rather than a loud failure.

## The open question, settled by evidence

#718 left `account_provider: simple_account` undecided — hard-error or accept-and-ignore. I checked avs-infra rather than guessing: **no config sets `account_provider` at all**, so every deployed chain takes the `modular_account_v2` default. Refusing `simple_account` therefore rejects a value nothing is using, and cannot break a deploy.

It is refused at **load**, not at send. A boot failure names the config line; a send-time failure names nothing and arrives per-operation, after users have already been handed legacy addresses.

## What changed

| | |
| --- | --- |
| `builder.go` (1,634 lines) | deleted |
| `waitForUserOpConfirmation` | **kept**, moved to `receipt_watch.go` — the one piece the v0.7 path uses, and EntryPoint-version agnostic |
| `SendUserOpAutoWithWsClient` | deleted — it existed to reuse the aggregator's long-lived socket, and the MA v2 path opens its own |
| `paymasterReq`, `executionFeeWei` | removed from the send signature — already documented as ignored on MA v2 |
| Paymaster probing at config load | removed — `owner()`/`verifyingSigner()` made boot depend on a live RPC call to a contract nothing invokes |
| `smart_wallet.paymaster_address` | still parses so existing configs load; logged once as ignored |

Falling out of the above: `shouldUsePaymaster` on both processors (it only gated the v0.6 request and already returned false on MA v2); the MAX-transfer `SkipReimbursement` case (no reimbursement leg to skip); and `DEFAULT_*_GAS` with the AA21 log block that compared against them — the v0.7 path seeds different values, so that comparison had stopped hiding "not yet estimated" placeholders and only obscured the gas figures it was meant to filter.

`ExecuteUserOpReq.UsePaymaster` is no longer set or read, but stays on the wire rather than becoming a proto break.

## Out of scope

v0.6 **derivation** and storage — `DefaultFactoryProxyAddressHex`, `aa.NewSimpleFactory`, the v0.6 factory on wallet records. Existing rows still reference it and `ProviderForFactory` still routes per wallet. This is the send path only.

## One thing to confirm before merging

`SendUserOpAuto`'s comment recorded an audit finding that pre-cutover v0.6 SimpleAccounts hold no meaningful balances. **I have not re-run it.** Those wallets already cannot execute on a cut-over chain, so this does not change their situation — it makes it permanent rather than reversible by config. Worth a second pair of eyes before merge.

## Test plan

- [x] `go build ./...`, `go vet ./...` clean
- [x] `core/config`, `pkg/erc4337/...`, `worker`, `aggregator/...` suites green
- [x] `TestSimpleAccountIsRefusedAtLoad` pins the refusal and that its message names the supported value
- [x] `TestValidateAccountProvider` updated — empty and `modular_account_v2` accepted, `simple_account` no longer
- [x] `go mod tidy` dropped four indirect deps only the removed code pulled in
- [ ] CI green — notably `Unit Test (pkg/erc4337/preset)` should now pass without depending on a funded paymaster


chrisli30:
Investigated the `TestResolveUserChainID` failure I waved off earlier. It is not that test, and it is not this PR — but my "flake, unrelated" call was asserted rather than demonstrated, so here is the evidence.

## It is a different test each run

Re-running the exact command that produced it — `go test ./core/... ./pkg/... ./worker/... ./aggregator/... -short` — failed again, on a **different** test at essentially the same duration:

| run | failing test | duration |
| --- | --- | --- |
| 1 | `TestResolveUserChainID` | 90.19s |
| 2 | `TestEventTriggerQueriesBasedConfiguration` | 90.71s |

A failure that moves between unrelated tests while holding its duration is a resource stall, not a defect in either test.

## The actual error

Run 2 surfaced it in full:

```
tenderly event simulation failed: context error before simulation: context deadline exceeded
```

An external-service call exceeding its deadline. Run 1's test does no Tenderly work but does derive a wallet address through `DeriveSenderAddressAuto` against the configured Sepolia RPC — the same shape of dependency, different provider.

The trigger is the sweep itself: `./core/... ./pkg/... ./worker/... ./aggregator/...` runs those packages **concurrently**, and they all reach the same free-tier endpoints (Tenderly, `ethereum-sepolia-rpc.publicnode.com`). Saturating them stalls whichever test happens to be holding a request.

## Why CI never sees it

`.github/workflows/run-test-on-pr.yml` runs one package per matrix job:

```yaml
cd ./${{ matrix.test }}
go test . -v -short
```

No cross-package concurrency, so no endpoint saturation. That is why `Unit Test (core/taskengine)` is green here and on #716 and #720.

## Ruled out as a cause of this PR

Timed directly: `TestResolveUserChainID` runs in **0.17s** whether or not a prior test has set the global `rpcConn` (I checked that specifically, since it changes which derivation branch `GetWallet` takes). It is 90s only under the sweep. And this PR touches neither Tenderly, event triggers, chain-ID resolution, nor wallet derivation.

## Worth noting separately

The sweep is a useful signal that the suite depends on shared free-tier external services to pass. That is not a blocker here, and not something to fix in this PR, but it means a local full-repo run is not a reliable gate — which is worth knowing before anyone treats one as such.

chrisli30:
Worked through the Copilot review — all three valid, two were defects this PR introduced. Fixed in 16d5918c.

**1. MAX-transfer gate still read the legacy paymaster address.** Right, and this one was a real bug. A MAX transfer sends the whole balance, so it only works if something else covers gas — now the Gas Manager policy, not `smart_wallet.paymaster_address`, which this PR made inert. Gating on the old field was wrong in *both* directions: a sponsored chain carrying no legacy address got refused, and a self-funded chain with a stale address was allowed to try and then failed on chain. The gate now reads `AlchemyPaymasterPolicyID`, and the ETH-transfer `shouldUsePaymaster` helper goes with it — reading the inert field was all it did. I had removed the contract-write copy and missed this one.

**2. `checkUserOpExecutionSuccess` kept its code but lost its test.** Right, and it contradicted this PR's own stated principle that removed tests correspond only to removed code. The function survived inside `receipt_watch.go` while its only coverage went out with `helpers_test.go`. Restored as `receipt_watch_test.go`, next to the code it covers, including the malformed-payload cases (short, empty, exactly-64-byte) that are the reason it exists.

**3. Worker path is not actually Gas Manager sponsored.** Correct on the facts, and my comment overclaimed — worker config carries no policy field for `ToSmartWalletConfig` to copy, so `priceOperationV07` never requests sponsorship and the operation takes the self-funded prefund path.

One correction: this is **not a regression from this PR**. The old MA v2 branch never received `paymasterReq` either —

```go
if smartWalletConfig.UsesModularAccountV2() {
    op, receipt, err := SendUserOpMAv2(smartWalletConfig, owner, callData, senderOverride, saltOverride, nil, lgr)
    // paymasterReq deliberately not threaded through; it was v0.6-only
}
```

— so worker-routed operations have been unsponsored since the v0.7 cutover. What this PR removed was the plumbing that made the path *look* sponsored. Wiring the policy through worker config is real work with its own decisions (per-worker vs shared policy, whether the webhook secret is needed), so it is tracked as #722 rather than bolted onto a deletion PR. The comment now states the situation plainly instead of describing behaviour the code does not have.

---

Also swept two things the review did not flag but this PR invalidated: `gateway.example.yaml` still documented `simple_account` as selectable and shipped a literal `paymaster_address`. Both corrected.

And per review feedback on endpoints, every `*.example.yaml` now warns against pointing `eth_rpc_url` at a free public endpoint — they rate-limit under load and fail runs nondeterministically on whichever request is in flight, which is exactly the ~90s "flake" investigated earlier in this PR.

PR #723, opened 2026-08-07
fix(worker): carry Gas Manager sponsorship into worker config

### LINKED ISSUE #722: Worker-routed UserOps are unsponsored: Gas Manager policy is not wired through worker config
## Summary

Operations sent through a chain worker are **not sponsored**. Gas Manager sponsorship requires `AlchemyPaymasterPolicyID` on the `SmartWalletConfig`, and worker config has no such field — so `ToSmartWalletConfig` cannot copy one, `priceOperationV07` never requests sponsorship, and the operation takes the self-funded prefund path. For a zero-balance smart wallet that fails.

Surfaced by the Copilot review on #721. **Not a regression from that PR** — see below — but that PR is what made it visible, because it removed the v0.6 paymaster plumbing that had been giving the path a misleading appearance of sponsorship.

## Why it is pre-existing

Before #721, `SendUserOpAuto` took a `paymasterReq` and the worker built one from `smart_wallet.paymaster_address`. But the MA v2 branch never passed it on:

```go
if smartWalletConfig.UsesModularAccountV2() {
    op, receipt, err := SendUserOpMAv2(smartWalletConfig, owner, callData, senderOverride, saltOverride, nil, lgr)
    // paymasterReq not threaded through — by design, it was v0.6-only
}
```

So worker-routed operations on MA v2 have been unsponsored since the EntryPoint v0.7 cutover. #721 removed the dead argument and, in doing so, removed the illusion.

## Scope

- `worker/config.go` (~13-34) exposes no paymaster policy id or webhook secret.
- `ToSmartWalletConfig` (~85-98) therefore cannot populate `AlchemyPaymasterPolicyID`.
- `aggregator/rpc_server.go` → `sendUserOpViaWorker` is the caller that lands here in gateway mode.

## What needs deciding

- **Where the policy id comes from.** Each worker has its own config file delivered through avs-infra, so the straightforward answer is a per-worker `alchemy_paymaster_policy_id` mirroring the gateway's. Confirm whether prod/staging workers should share the gateway's policy or hold their own.
- **Whether the webhook secret is needed too.** If sponsorship uses the gas-manager webhook, the worker needs that secret as well, which is a different blast radius than a policy id.
- **Whether unsponsored should be loud.** Right now a worker-routed operation for a zero-balance wallet fails at prefund with an AA-level error. Refusing earlier, with a message naming the missing config, would be easier to diagnose.

## Acceptance criteria

- [ ] A worker-routed UserOp is sponsored by the same Gas Manager policy a gateway-routed one would use.
- [ ] Worker config carries the policy explicitly rather than inheriting by accident.
- [ ] A worker configured without a policy either sponsors correctly or fails with a message that names the missing field — not an opaque prefund failure.
- [ ] Coverage that distinguishes sponsored from self-funded on the worker path, so this cannot silently regress again.

## Context

| Resource | Note |
| --- | --- |
| The comment recording this | `worker/server.go`, in `ExecuteUserOp` |
| Sponsorship request | `priceOperationV07`, gated on `AlchemyPaymasterPolicyID != ""` |
| Gateway equivalent | `smart_wallet.alchemy_paymaster_policy_id` in `config/gateway.example.yaml` |
| Bundler/paymaster provider switch | [[project_bundler_provider_flag]] — alchemy is the default |

chrisli30:
Closes #722.

## Summary

Operations sent through a chain worker were **unsponsored**. `priceOperationV07` only asks Alchemy to sponsor when `AlchemyPaymasterPolicyID` is set on the `SmartWalletConfig`, and worker config had no such field for `ToSmartWalletConfig` to copy — so every worker-sent operation fell through to `ensureSelfFundedPrefund`, which a smart wallet holding only tokens cannot satisfy.

## What it actually affected

I traced the blast radius rather than assuming it was everything. The worker is a chain-RPC proxy: reads, plus exactly one write path, `ExecuteUserOp`. That has exactly one caller — `sendUserOpWithGlobalWs` — which itself has exactly one caller: **`ExecuteWithdraw`**.

So task execution was fine (it runs in the gateway process against the gateway's config, which does carry the policy). What was broken is **withdrawal — the escape hatch**. A wallet holding USDC but no ETH could not pay its own prefund, so the user could not get their tokens out.

**Not a regression from #721.** The MA v2 branch never received the old `paymasterReq` either:

```go
if smartWalletConfig.UsesModularAccountV2() {
    op, receipt, err := SendUserOpMAv2(smartWalletConfig, owner, callData, senderOverride, saltOverride, nil, lgr)
    // paymasterReq deliberately not threaded through; it was v0.6-only
}
```

This has been true since the EntryPoint v0.7 cutover. Removing the v0.6 plumbing only stopped the path from *looking* sponsored.

## The non-obvious half

The worker also needs `gas_manager_webhook_secret`, which reads like gateway-only config. The webhook endpoint does stay on the gateway (`POST /webhooks/gas-manager`) — but the secret is echoed to Alchemy as `webhookData`, and the gateway's webhook rejects any request whose value does not match. A worker that requests sponsorship without it gets refused by the policy's custom rules.

## Approach

Resolution moves into exported `config.ResolveAlchemyPaymasterPolicyID` / `ResolveGasManagerWebhookSecret`, now used by **both** gateway and worker. The bug was the two disagreeing about where sponsorship comes from; one function is harder to drift than two copies. Legacy `gas_manager_policy_id` and the env fallbacks are preserved exactly, so no existing deployment silently loses sponsorship.

A worker with no policy now says so at startup. Deliberately a **warning, not a refusal** — a chain may legitimately run self-funded, and the send path already fails with a message naming `alchemy_paymaster_policy_id` when a wallet cannot cover prefund. That existing error is good, so I did not add a second one.

## Deployment — this needs an avs-infra change

The code cannot fix this alone. Worker configs in `avs-infra/railway/configs/` currently carry **no** `alchemy_paymaster_policy_id` — the key appears only in `gateway-railway.yaml`. Each worker needs both fields set to match the gateway before sponsorship actually resumes.

Open question for whoever does that: whether workers share the gateway's policy or hold their own. Sharing is simpler and matches how the gateway is configured today; separate policies would give per-chain spend caps. I have not assumed either — the config reads whatever is supplied.

## Test plan

- [x] Policy and webhook secret reach `SmartWalletConfig`
- [x] A worker without a policy reports it cannot sponsor (the regression itself)
- [x] Legacy `gas_manager_policy_id` alias accepted; canonical wins when both set
- [x] Env fallback fills in when yaml is empty
- [x] Whitespace trimmed — a padded policy id reads to Alchemy as an unknown policy
- [x] `go build`, `go vet`, `gofmt`, `make storage-check` clean
- [ ] CI green
- [ ] avs-infra: set both fields on every worker, then confirm a withdrawal from a zero-ETH wallet succeeds

### DIFF
--- a/aggregator/task_engine.go
+++ b/aggregator/task_engine.go
@@ -75,7 +75,7 @@ func (agg *Aggregator) startTaskEngine(ctx context.Context) {
 			"effective_factory", effectiveFactory.Hex(),
 			"config_factory_address", sw.FactoryAddress.Hex(),
 			"entry_point", sw.EntryPointAddress().Hex(),
-			"paymaster_policy_set", sw.AlchemyPaymasterPolicyID != "",
+			"paymaster_policy_set", sw.SponsorshipPolicyID() != "",
 			"legacy_v06_paymaster", sw.PaymasterAddress.Hex(),
 		)
 	}
--- a/config/gateway.example.yaml
+++ b/config/gateway.example.yaml
@@ -187,3 +187,15 @@ notifications:
     provider:     context-memory
     api_endpoint: "http://localhost:3010"
     api_key:      ""
+
+# Gas Manager sponsorship is DISABLED for local/development runs, deliberately.
+#
+# The policy's custom-rules webhook is a single URL pointing at the PRODUCTION
+# gateway, so a locally-run process that requests sponsorship has production
+# approve it — deciding against production's wallet records and billing
+# production's budget for your machine. Leave this true locally and fund the
+# test smart wallet with the chain's native token instead.
+#
+# It is an explicit setting rather than an inference from `environment` so that
+# changing log verbosity can never change who pays for gas.
+disable_gas_sponsorship: true
--- a/config/test.example.yaml
+++ b/config/test.example.yaml
@@ -118,3 +118,15 @@ notifications:
     provider:     context-memory
     api_endpoint: "http://localhost:3010"
     api_key:      ""
+
+# Gas Manager sponsorship is DISABLED for local/development runs, deliberately.
+#
+# The policy's custom-rules webhook is a single URL pointing at the PRODUCTION
+# gateway, so a locally-run process that requests sponsorship has production
+# approve it — deciding against production's wallet records and billing
+# production's budget for your machine. Leave this true locally and fund the
+# test smart wallet with the chain's native token instead.
+#
+# It is an explicit setting rather than an inference from `environment` so that
+# changing log verbosity can never change who pays for gas.
+disable_gas_sponsorship: true
--- a/config/worker-base-sepolia.example.yaml
+++ b/config/worker-base-sepolia.example.yaml
@@ -57,3 +57,15 @@ gas_manager_webhook_secret:  ${GAS_MANAGER_WEBHOOK_SECRET}
 smart_wallet:
   controller_private_key: <testnet-controller-private-key-hex>
   paymaster_address:      0xd856f532F7C032e6b30d76F19187F25A068D6d92
+
+# Gas Manager sponsorship is DISABLED for local/development runs, deliberately.
+#
+# The policy's custom-rules webhook is a single URL pointing at the PRODUCTION
+# gateway, so a locally-run process that requests sponsorship has production
+# approve it — deciding against production's wallet records and billing
+# production's budget for your machine. Leave this true locally and fund the
+# test smart wallet with the chain's native token instead.
+#
+# It is an explicit setting rather than an inference from `environment` so that
+# changing log verbosity can never change who pays for gas.
+disable_gas_sponsorship: true
--- a/config/worker-base-sepolia.example.yaml
+++ b/config/worker-base-sepolia.example.yaml
@@ -39,6 +39,20 @@ bundler_url:      ${BASE_SEPOLIA_BUNDLER_URL}
 bundler_provider: self_hosted
 alchemy_api_key:  ${ALCHEMY_API_KEY}
 
+# Gas Manager sponsorship — MUST match the gateway, or this worker sends
+# UNSPONSORED operations.
+#
+# Sponsorship is requested only when a policy id is present. Without one, every
+# operation this worker sends has to be paid for out of the smart wallet's own
+# balance, so a wallet holding only tokens cannot move them — which is how
+# withdrawals routed through a worker silently stopped being sponsored (#722).
+#
+# The webhook endpoint itself stays on the gateway (POST /webhooks/gas-manager).
+# What the worker needs is the SECRET: it is echoed to Alchemy as webhookData,
+# and the gateway's webhook rejects any request whose value does not match.
+alchemy_paymaster_policy_id: ${ALCHEMY_PAYMASTER_POLICY_ID}
+gas_manager_webhook_secret:  ${GAS_MANAGER_WEBHOOK_SECRET}
+
 # Smart wallet / account abstraction config.
 smart_wallet:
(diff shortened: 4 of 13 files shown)


chrisli30:
Worked through the Copilot review — all three were real, and the first would have broken a chain that works today.

**1. Self-hosted bundler cannot serve a sponsorship request.** The important one, and I had not considered it. Sponsorship goes out as `alchemy_requestGasAndPaymasterAndData` against `ActiveBundlerURL`, which Voltaire does not implement.

Checking production rather than the templates: **4 of 5 workers are `bundler_provider: alchemy`, but `worker-bnb-mainnet` is `self_hosted`** — as is chain 56 inside the gateway's own chain list. So handing every worker the shared policy would have made BNB attempt sponsorship and fail the send, where today it runs self-funded and works. That is a regression this PR would have introduced, on the one chain that differs from the rest.

A policy is now inert unless the chain's bundler is Alchemy.

**2. `environment` is a logging setting, not an authorization switch.** Agreed, and I should not have reached for it. `ConfigRaw.Environment` is `sdklogging.LogLevel` and selects Zap formatting; overloading it meant that raising log verbosity in production would silently stop sponsoring — a logging change moving money. It also made the templates self-contradictory, documenting `alchemy_paymaster_policy_id` next to `environment: development` where it could never take effect.

Replaced with `disable_gas_sponsorship`, which says what it does. The local templates set it, so the original point still holds: a stray exported `ALCHEMY_PAYMASTER_POLICY_ID` cannot opt a laptop into the production policy.

**3. Trimming after `firstNonEmpty`.** Correct — a yaml value of `" "` is non-empty, so it beat a valid environment fallback and then trimmed to empty, silently unsponsored. Both resolvers now trim each candidate before selecting, and there is a test for the whitespace case specifically.

---

All three conditions land behind `SmartWalletConfig.SponsorshipPolicyID()`, which every sponsorship decision now reads instead of `AlchemyPaymasterPolicyID` directly — six call sites across send, contract-write, ETH-transfer, the webhook and startup logging. Conditions like these do not stay applied consistently when they are spread out; that is the same reasoning that put resolution in one shared function earlier in this PR.

Startup logging now distinguishes all three states: opted out, no policy, or policy present but unusable on this bundler (naming the provider).

CI green. One deployment note carried forward: BNB will remain self-funded until its bundler moves to Alchemy — worth deciding separately whether that is wanted, since it is now explicit rather than accidental.

chrisli30:
Second Copilot pass: **no new comments**, and two suppressed ones — both making the same point about `worker/worker.go`. Acted on them anyway in 1b2e7e36, because the half Copilot led with was the cheaper half.

Startup logging called `WorkerConfig.SponsorshipConfigured()`, which rebuilt a `SmartWalletConfig` — re-parsing the controller key — even though `NewWorker` already builds one and stores it. Copilot framed that as duplicate parsing, which it is. The part worth more: it was a **second way to answer "will this worker sponsor?"**, and two answers to one question is exactly the shape of the bug this PR exists to fix. The gateway and worker disagreeing about where sponsorship comes from is how worker-routed operations ended up unsponsored in the first place.

Startup now reads `w.smartWalletCfg` — the config `SendUserOpMAv2` is actually handed — so the log cannot report something the send path will not do.

`SponsorshipConfigured` is **deleted** rather than left unused. Once startup stopped calling it, its only remaining callers were the tests asserting on it: a predicate kept alive by its own coverage. The tests now resolve sponsorship through `ToSmartWalletConfig().SponsorshipPolicyID()`, the same path production takes, so they can no longer pass while the real answer differs.

Full suites green, `make storage-check` clean.

PR #731, opened 2026-08-07
feat(taskengine): complete session-grant on-chain replace/revoke lifecycle (#717)

### LINKED ISSUE #717: Session grants: batch uninstallValidation into the deferred action so revoke/replace reaches the chain

chrisli30:
## Summary

**Closes #717.** Completes on-chain session-grant lifecycle for replace and explicit revoke.

### Review fixes (Copilot + Claude on this PR)

| Finding | Fix |
| --- | --- |
| Forever-growing teardown set / stale GET cleanup | `Grant.TornDownAt` set when replace teardown verifies clear; targets + GET gated by `NeedsOnChainCleanup()` |
| N-way under-seeds verification gas | `DeferredTeardownCount * 120_000` added to deferred verification seed (~98k measured + headroom) |
| Cap 16 hard-refuses prepare | Soft cap **4** — take first N, never refuse; leftovers ride next grant |
| OpenAPI pending wording | DELETE docs: pending with InstallCall → `revoked`, not `deleted` |
| entityId int32 overflow | OpenAPI `int64` |
| Dead `OnChainCleanupForPolicy` | Removed |
| Named-return shadowing | Fixed |
| Weak unwrap test | Asserts entity 2 vs 1 |

### Lifecycle

| Gap | Fix |
| --- | --- |
| **AC1 — one installed entity after replace** | N-way deferred batch (≤4 teardowns/op) + TornDownAt so history is not re-uninstalled |
| **AC3 — explicit revoke** | `onChainCleanup` on DELETE; GET only when still needs cleanup |
| **Orphans** | Retain pending as revoked; late MarkApplied sets AppliedAt without usable resurrection |
| **Failure semantics** | `SESSION_GRANT_INSTALL_FAILED` → client Warn |

### Explicit revoke (owner still signs the chain call)

Production grants are policied: controller cannot self-uninstall (R4). Gateway returns payload; Studio/wallet executes.

### Not in this PR

- Live Sepolia proof for **N≥2** teardown (unit tests pack N-way + gas seed; 1-way live remains from #729). Optional follow-up with stacked fixture.
- Studio wiring of `onChainCleanup` → `sendTransaction`.

## Test plan

- [x] Unit: TornDownAt excludes from batch/cleanup; N-way gas seed; unwrap entity id
- [x] `go test` taskengine, aa, preset, rest, model
- [x] `make storage-check` — TornDownAt additive only
- [ ] CI green
- [ ] Optional: live N=2 replace on Sepolia


chrisli30:
## Audit

Reviewed against #717's acceptance criteria, built and ran the suites (`taskengine`, `aa`, `preset`, `rest` — all green), and measured the on-chain cost of the composition this PR generalises.

The shape is right: N-way teardown derived from stored `InstallCall`, orphans retained rather than deleted, cleanup payload built by the same function the batch uses. Two findings would keep me from merging as-is, both concentrated on the N-way path.

---

### 1. The N-way batch outgrows the verification-gas seed, which this PR does not touch

A deferred action executes **during validation**, so every uninstall in the batch is charged against `verificationGasLimit`, not call gas.

Measured on Sepolia from the #729 fixture runner (`0x0643f9d156e2CE584825511C548649F3289619c3`), reading `UserOperationEvent.actualGasUsed` straight off the EntryPoint:

| operation | entities | actualGasUsed |
| --- | --- | --- |
| plain install, no teardown | 3, 5, 33 | 546,745 |
| install + **one** uninstall | 2, 4, 34 | 644,715 |

**≈ 97,970 gas per teardown**, all of it inside validation.

`seedVerificationGasFor` returns a flat `seedVerificationGasDeferredHooks = 700_000` (`pkg/erc4337/preset/builder_v07.go:78`) and takes no account of how many uninstalls the deferred call carries. `session_auth.go` is not in this PR's diff. At the cap of 16 that is **+1,567,520 gas** of validation work against an unchanged 700k seed.

This is not recoverable at estimation. `EstimateUserOpGasV07` sends the seeded `verificationGasLimit` in the `eth_estimateUserOperationGas` payload, and the batch runs inside that simulation — the code says so directly: *"over-seeding is corrected by SendUserOpV07WithRetry's efficiency tighten, under-seeding is AA23/AA26 at estimation."* The tighten only ever corrects downward.

Working from ~546k for install-only and ~98k per teardown, the seed is exhausted somewhere around **N = 3–4**. Whatever the exact number, the consequence is the same and it points the wrong way:

- Before this PR, a stacked wallet took the `ambiguous` branch, installed only, and **could still be granted**. #716's self-healing held.
- After it, that wallet builds an N-way batch that fails estimation and **cannot be granted at all** — the exact wallet class AC1 exists to fix.
- And `wrapDeferredGrantInstallError` renames the resulting AA23 to *"deferred grant install/replace did not land"*, so the operator sees a symptom that reads like a chain problem rather than an over-large batch.

Suggested fix: seed from the deferred call — count the uninstalls in it and add the measured ~100k each — and derive `maxOnChainTeardowns` from what the seed can actually cover rather than 16. If a hard cap stays, worth stating what it was measured against.

### 2. Nothing proves N ≥ 2 on chain, and the live test was edited to keep it that way

The table claims AC5 via #729's live test. That test asserts exactly one supersede, and this PR changes it to keep asserting exactly one:

```go
require.Len(t, prepared.Supersedes, 1, "an installed grant must be superseded on chain too")
```

So the only end-to-end evidence still covers the 1-way case that R-A already proved. "N-way is the same composition with more uninstalls" is plausible and is the part I would least want to take on faith here — the recurring lesson in this area is that these batches mine, report `success = true`, and clear nothing, which is why `aa.EntitySignerOnChain` and `VerifySupersededTeardown` exist at all. Finding 1 suggests N ≥ 2 may not reach the chain at all today.

The fixture makes this cheap: salt 13 currently carries four stacked installed entities (2, 4, 5, 6) — a natural pathological wallet. Seeding storage with two applied policies and running the existing live test would settle findings 1 and 2 together.

---

### 3. `> 16` hard-refuses prepare, which reverses a deliberate decision

`onChainTeardownTargets` now returns an error above the cap, so a wallet with 17+ applied entities cannot be granted at all and the suggested remedy is 17 owner transactions. The comment this PR deletes argued the opposite case explicitly — *"refusing here would take that away precisely from the wallets that need it most"*. Bounded and arguably fine, but it is a reversal and reads here like a side effect of adding the cap rather than a decision. Degrading to install-only above the cap would keep self-healing and still bound the batch.

### 4. `policyToAPI` silently swallows the cleanup-build failure

```go
if c, err := taskengine.BuildOnChainRevokeCleanup(p); err == nil {
    out.OnChainCleanup = onChainCleanupAPI(c)
}
```

On error the field is omitted and the client sees a revoked+applied grant with no cleanup payload and no reason — indistinguishable from one that needs none. Same class of thing `aa.EntitySignerOnChain` documents against: *"A read failure is returned rather than folded into the zero address."* A log line would be enough; the response shape can stay.

### 5. Minor

- **`OnChainCleanupForPolicy` has no callers.** Its comment says "Used by GET", but GET renders through `policyToAPI`, which builds the payload inline. Either wire it up or drop it.
- **Shadowed named return** in `RevokeSessionPolicyByID`: `cleanup, buildErr := BuildOnChainRevokeCleanup(policy)` shadows the named `cleanup`. Correct today because the explicit `return` names the inner one, but the next edit to that block will not be.
- `TestBuildOnChainRevokeCleanupUnwrapsAReplaceBatch` asserts `cleanup.CallData == aa.SessionSignerUninstallFromInstall(2, batch)` — the same call the implementation makes, so it passes even if the wrong install were peeled out of the batch. Asserting the entity actually torn down is `2` and not `1` would close that.

---

### Worth keeping, and worth saying why

Retaining a pending grant with an `InstallCall` as revoked instead of deleting it is the right call, and it is load-bearing beyond the orphan story: `NextSessionEntityID` allocates `max(entityID)+1` across **all** retained records, and a deleted record lets a spent entity be reissued. A torn-down entity reads clear on every module while its EntryPoint deferred nonce sequence never resets, so a reissued entity signs a carrier nonce the EntryPoint has already consumed and AA23s with nothing on chain to explain it (see `aa.EntryPointNonce`, #729). Deleting those records would have reopened exactly that. Might be worth a sentence in the doc comment, since the connection is not obvious from the orphan framing alone.

Deriving both the batch and the revoke payload from stored `InstallCall` through the single `SessionSignerUninstallFromInstall` path is also the right structure — that is what keeps a grant whose shape predates a code change tearing down correctly.

chrisli30:
## Re-audit of `6c6c3383`

Rebuilt, reran `taskengine` / `aa` / `preset` / `rest` (green), and re-checked storage: `SessionGrantAuthorization.TornDownAt added (int64) — backward-compatible`, no key-template changes.

**Round-1 items resolved.** The gas seed now scales, and it is wired from the right source — `DeferredTeardownCount` comes from `aa.UninstalledEntitiesWithin(policy.Grant.InstallCall)`, i.e. counted off the payload the owner actually signed rather than a separately stored number that could drift. Against the measurements, it holds at every N up to the new cap:

| N | seed | measured total op gas |
| --- | --- | --- |
| 0 | 700,000 | 546,745 |
| 1 | 820,000 | 644,715 |
| 4 | 1,180,000 | 938,625 |

(measured totals include preVerification and call gas, so the verification-only figure sits further under each seed, and efficiency stays well clear of the 0.4 floor). Cap 4 with soft-degrade restores the self-healing property. Dead code gone, shadowed return renamed, and the batch-unwrap test now asserts the entity rather than re-running the implementation.

`TornDownAt` is a good addition and solves a real problem I had not raised — without it every historical applied grant stays a teardown target forever.

That last one, though, is where this now breaks.

---

### Blocking — using the new cleanup payload makes the wallet ungrantable

**Re-uninstalling an already-cleared entity reverts.** Proven on Sepolia against the #729 fixture runner (`0x0643f9d156e2CE584825511C548649F3289619c3`), `eth_call` with `from` = the account so it takes the same self-call path `executeBatch` produces, identical payload construction for both, only the on-chain state differs:

```
entity 1  CLEARED (prior replace)   uninstall REVERTS: execution reverted
entity 2  INSTALLED                 uninstall succeeds
```

Now follow the explicit-revoke path this PR adds:

1. Owner revokes an applied grant → gets `onChainCleanup` → executes it in their wallet. The entity is cleared on chain.
2. The gateway never learns. `TornDownAt` is written in exactly one place — `MarkEntityTornDown`, reached only from `VerifySupersededTeardown` in the replace flow's `OnApplied`. There is no confirm endpoint and no readback on the revoke path.
3. Owner grants again. `onChainTeardownTargets` still sees `Applied() && !TornDown()`, so the batch is `install(new) + uninstall(cleared entity)`.
4. That uninstall reverts → the whole deferred batch reverts → **the new grant never installs**.

And it does not recover on its own: the only thing that sets `TornDownAt` is a *successful* replace, which is precisely what is now failing. Re-revoking does not help — `RevokeSessionGrant` on an already-revoked applied policy just re-stores it. The wallet is stuck until someone edits storage.

So AC3's headline feature — hand the owner a payload and let them finish teardown — is what triggers it. The more diligent the user, the sooner they hit it.

The same mechanism fires on the replace path whenever verification does not complete: `verifyTeardown == nil` (no RPC configured for that chain), a timeout, or a transient read failure all leave `TornDownAt` unset. The comment at `session_grant.go:589` reads *"a failed mark only means the next grant may re-attempt uninstall (harmless if already clear)"* — that is the assumption the probe above falsifies. It is not harmless; it is fatal, and it makes a transient RPC blip permanently break granting for that wallet.

Suggested fix: do not let storage belief alone decide what goes in the batch. `TeardownVerifier` already exists and the engine already wires it — reading each candidate's signer in `onChainTeardownTargets` (or immediately before packing) and dropping the ones already clear would close both the explicit-cleanup case and the failed-verification case in one move, and would set `TornDownAt` as a side effect. That is the same rule this subsystem already lives by everywhere else: a mined transaction is not evidence, so read the chain.

### Still open — nothing proves N ≥ 2 on chain

Unchanged from the last round: `session_grant_replace_live_test.go` still asserts `require.Len(prepared.Supersedes, 1)`, so the only end-to-end evidence covers the 1-way case. The seed arithmetic above is convincing on paper and I have no reason to doubt it, but the finding above is exactly the class of thing that arithmetic does not catch — it took an `eth_call` against real state to see it. Salt 13 currently carries stacked entities (2, 4, 5, 6 installed; 1, 3 cleared), which is both the N-way fixture and, conveniently, a wallet that already contains cleared entities to regression-test the fix against.

### Moderate — `MarkEntityTornDown` writes outside the runner lock

`ListSessionPolicies` → mutate → `StoreSessionPolicy` with no `sessionAuthorityLock` (the only one in the file is at line 343, inside `MarkSessionGrantAppliedByID`). That is the window the lock exists to close, described in the comment this PR edited: *"a submit that supersedes this grant between the read and the store would be overwritten by the store."* Concretely, a concurrent `MarkSessionGrantAppliedByID` on the same record loses its `AppliedUserOpHash` write.

Taking the lock here is safe — `OnApplied` calls `MarkSessionGrantAppliedByID` first and it has released by the time `VerifySupersededTeardown` runs, so there is no nesting.

### Minor — the stranded-entity log names the wrong policy

The resolver builds `replaced` with `ID: policyID`, which is the **new** grant's id, for every superseded entity. So the one log line that matters when this goes wrong —

> `a replaced grant is still installed on chain after its teardown mined  policy=<id> entity=<n>`

— names the grant that is fine, not the one that is stranded. `MarkEntityTornDown` keys off `(runner, entity)` so the write still lands correctly; this is diagnostics only, but it is the diagnostics for the failure mode this subsystem exists to catch. Looking the id up from the entity, or leaving it empty, would beat naming the wrong one.

PR #732, opened 2026-08-07
test(taskengine): live Sepolia proof for N-way teardown and the cleared-candidate drop

chrisli30:
Live Sepolia coverage for the two behaviours #731 shipped that had never run against a chain.

#731 generalised the replace batch from one prior entity to N, and added a chain read that drops candidates already clear. Neither had live coverage — `TestGrantReplaceClearsThePriorEntity_Sepolia` asserts exactly one supersede, so it passes unchanged whether or not N-way works.

Both failures are invisible offline:

- A misrouted uninstall **mines with `success = true` and clears nothing** — the account catches `onUninstall` reverts and strands the module state.
- An uninstall aimed at an **already-clear** entity **reverts**, taking the whole deferred batch with it, so the new grant never installs.

The second is what `dropClearedTeardownTargets` exists for. Storage keeps believing an entity is installed long after the owner cleared it through the `onChainCleanup` payload, because nothing tells the gateway it happened — and only a *successful* replace sets `TornDownAt`, so without the drop the wallet cannot recover by itself.

## What the test does

Stacks two entities through the production path, so both `InstallCall`s are the bytes the owner actually signed rather than reconstructions — teardown is derived from those, and a rebuilt payload would mine and clear nothing. Then:

1. one grant tears down both, and the modules are read back;
2. the post-cleanup state is reproduced (entity clear on chain, storage still believing it installed) and the stale candidate must be **dropped**, recorded `TornDownAt`, and the grant must still land.

```
stacked: entities 25 and 26 both installed
✅ entities 25 and 26 cleared in one batch; 27 holds the replacement
✅ stale candidate 25 dropped; grant 28 still landed
```

## Verified to catch what it is for

| control | result |
| --- | --- |
| `dropClearedTeardownTargets` disabled | fails on the stale candidate — *"entity 11 is already clear on chain; batching its uninstall reverts the whole grant"* |
| `maxOnChainTeardowns` forced to 1 | fails on the supersede count — *"both installed entities must ride this batch"* |

## Also fixes a latent flake in the existing live test

`ValidUntil` was recomputed from `time.Now()` on the way into submit. Submit does not trust the payload it is handed — it re-runs `PrepareSessionGrant` and verifies the owner's signature against the digest it rebuilds — and `ValidUntil` reaches the TimeRange hook as unix **seconds**. A prepare and submit landing either side of a second boundary therefore rebuild a different digest, and the signature recovers to a stranger:

```
grant was signed by the wrong key: signed by 0x0609448b…, not the wallet owner 0x804e49e8…
```

Harmless until #731 added a per-candidate chain read to prepare (`teardownVerifier` dials per check), which pushed prepare into the hundreds of milliseconds and made the crossing routine — it cost two runs here before I stopped blaming the product. Now pinned for the process.

Worth noting for its own sake: that error message sends you to key management when the actual cause is an input that drifted between prepare and submit. A real client sends the same `validUntil` it prepared with, so this is not a live defect, but the diagnostic would be misleading if one ever did drift.

## Notes

- Behind the `integration` build tag; unit CI is unaffected.
- Runs on the salt-13 fixture and consumes four validation entities per run, which can never be recycled (#719) — on-demand, not per-PR.
- `scripts/fixture_wallet -salt 13` reports the wallet's per-entity state and funds it.

Refs #717

### DIFF
--- a/core/taskengine/session_grant_nway_live_test.go
+++ b/core/taskengine/session_grant_nway_live_test.go
@@ -146,18 +146,27 @@ func TestGrantReplaceClearsSeveralPriorEntities_Sepolia(t *testing.T) {
 	clearTornDownMark(t, db, chainID.Int64(), owner, *runner, first.EntityID)
 
 	afterCleanup := firstPrepared(t, engine, user)
-	for _, s := range afterCleanup.Supersedes {
-		require.NotEqual(t, first.EntityID, s.EntityID,
-			"entity %d is already clear on chain; batching its uninstall reverts the whole grant",
-			first.EntityID)
-	}
+
+	// Pin the whole set, not just the absence of the stale one. Dropping every
+	// candidate would also satisfy "the cleared entity is gone" while quietly
+	// letting the new grant install over a live one — the stacking this exists
+	// to prevent. Exactly one entity is installed at this point (third), so
+	// exactly one must ride the batch.
+	require.Len(t, afterCleanup.Supersedes, 1,
+		"only the already-clear entity may drop out; the installed grant must still be torn down")
+	require.Equal(t, third.EntityID, afterCleanup.Supersedes[0].EntityID,
+		"the batch must supersede the entity that is actually installed")
 	require.True(t, tornDownOnRecord(t, db, chainID.Int64(), owner, *runner, first.EntityID),
 		"a candidate found clear on chain must be recorded torn down, or every later grant re-reads it")
 
 	fourth := submitPrepared(t, engine, user, ownerKey, afterCleanup)
 	runOperationUnderGrant(t, engine, user, *runner)
+
 	requireEntityClear(t, client, *runner, fourth.EntityID, false)
-	t.Logf("✅ stale candidate %d dropped; grant %d still landed", first.EntityID, fourth.EntityID)
+	requireEntityClear(t, client, *runner, third.EntityID, true)
+	requireEntityClear(t, client, *runner, first.EntityID, true)
+	t.Logf("✅ stale candidate %d dropped, live entity %d still torn down; grant %d landed",
+		first.EntityID, third.EntityID, fourth.EntityID)
 }
 
 // clearTornDownMark removes the TornDownAt marker from whichever policy holds
--- a/core/taskengine/session_grant_nway_live_test.go
+++ b/core/taskengine/session_grant_nway_live_test.go
@@ -1,0 +1,197 @@
+//go:build integration
+// +build integration
+
+package taskengine
+
+import (
+	"context"
+	"math/big"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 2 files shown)


chrisli30:
Both addressed in `475521c5`.

**Step 3 assertion — you were right, and the hole was reachable.** It asserted the already-clear entity was *absent* from `Supersedes` but never that the installed one was *present*, so an empty `Supersedes` satisfied a `NotEqual` loop over zero elements. Now the whole set is pinned — exactly one entry, and it must be the entity that is actually installed — and the chain is read back for the live entity too, not just the replacement.

Verified with a control that isolates step 3 specifically: drop every candidate whenever any one of them is already clear. Phase 2 has no cleared candidate, so it passes untouched; phase 3 fails on the new assertion.

```
✅ entities 43 and 44 cleared in one batch; 45 holds the replacement     ← phase 2 unaffected
session_grant_nway_live_test.go:155:
    Error: "[]" should have 1 item(s), but has 0
    Messages: only the already-clear entity may drop out; the installed grant must still be torn down
```

The old loop passed that mutation.

**A second bug surfaced while re-running.** The entity scan in `seedEntitiesConsumedOnChain` was bounded at a constant `32`, and the fixture wallet outgrew it. The scan then reported a highest that was too low, the allocator handed out entity 33 — clear on every module, deferred nonce sequence already `1` — and every grant AA23'd during validation with nothing on chain to explain it.

That is precisely the failure mode `aa.EntryPointNonce` documents, walked into by the helper written to prevent it. It cost three runs and a gas-seed experiment before I checked the sequences above the bound:

```
entity 32  seq=1  installed
entity 33  seq=1  SPENT (cleared but nonce used)   ← allocator handed this out
entity 34  seq=1  installed
entity 35  seq=0  free
```

The scan now walks until a run of consecutive free entities — a single gap does not stop it, since a torn-down entity between two live ones is normal after a replace — with a ceiling that fails loudly and says to rotate the salt.

Both live tests pass on the current fixture:

```
✅ entity 35 cleared, entity 36 holds the replacement
✅ entities 37 and 38 cleared in one batch; 39 holds the replacement
✅ stale candidate 37 dropped, live entity 39 still torn down; grant 40 landed
```

chrisli30:
Taken — `475521c5`'s follow-up.

The reasoning holds: a non-zero deferred nonce sequence means the entity can never be reissued, so the signer read adds nothing in that branch. It also matters more now than when the suggestion was written, because the same review round changed this scan from a fixed 32 iterations to a walk that ends on a run of consecutive free entities — so its cost grows with the fixture rather than staying constant. At the current depth that is **106 calls down to 61**.

One thing I did not do is drop the signer read entirely. A zero sequence is not the same as free: an entity installed by anything other than a deferred action — an owner calling `installValidation` directly, which is the inverse of the explicit-cleanup flow this PR tests — holds a signer while its deferred nonce key is untouched. Removing the read would report such an entity as free and hand it to the next grant. So it stays for the zero-sequence case, with a comment saying why.

Both live tests re-run green against the same fixture, reporting the same spent boundary:

```
entities 1..45 are already spent on chain; allocating above them
✅ entity 46 cleared, entity 47 holds the replacement

entities 1..47 are already spent on chain; allocating above them
✅ entities 48 and 49 cleared in one batch; 50 holds the replacement
✅ stale candidate 48 dropped, live entity 50 still torn down; grant 51 landed
```

PR #737, opened 2026-08-07
feat: session-grant on-chain lifecycle, fixture isolation, CI coverage

chrisli30:
- **feat(taskengine): complete session-grant on-chain replace/revoke lifecycle (#717) (#731)**

Co-authored-by: Chris Li <chris@avaprotocol.org>

- **fix(worker): stack traces at Error, not Warn; restore CI for 12 skipped packages (#735)**

Co-authored-by: Chris Li <chris@avaprotocol.org>

- **test(taskengine): live Sepolia proof for N-way teardown and the cleared-candidate drop (#732)**

Co-authored-by: Chris Li <chris@avaprotocol.org>

- **test(taskengine): give each writing live fixture its own wallet (#719) (#733)**

Co-authored-by: Chris Li <chris@avaprotocol.org>


chrisli30:
All six findings addressed in `5caeb04e`. Summary first, then the one that mattered.

| # | Finding | Status |
| --- | --- | --- |
| Copilot 1 | `MarkEntityTornDown` retakes the runner lock under submit → deadlock | **FIXED** |
| Copilot 2 | `-fund-usdc` accepts non-positive amounts | **FIXED** |
| Claude 1 | Fixed 10s budget for up to 4 sequential dial-and-reads | **FIXED** |
| Claude 2 | Cap applied before the cleared-entity filter | **FIXED** |
| Claude 3 | `PrepareSessionPolicy` doc claims it stores nothing | **FIXED** |
| Claude 4 | `deleted` inferred rather than reported | **FIXED** |

## The deadlock is real, and it is mine

`SubmitSessionPolicy` holds the runner's write lock across `PrepareSessionGrant` → `dropClearedTeardownTargets` → `MarkEntityTornDown`, which took the same non-reentrant mutex. That wedges the `(chain, owner, runner)` shard for the life of the process — every later grant, revoke and confirmation for that wallet blocks behind it.

I asked for that lock in the #731 re-audit and checked only the `OnApplied` caller for nesting, missing that the same commit added a second caller already inside it.

It hides well: prepare normally marks the entity first, so submit's re-prepare finds nothing left to mark. The window is an entity cleared **between** prepare and submit — which is precisely what an owner running the `onChainCleanup` payload does, i.e. the flow #731 added.

Pinned by a regression test that stands where submit stands and fails by timeout without the fix:

```
--- FAIL: TestDropClearedTeardownTargetsDoesNotRetakeTheRunnerLock (5.05s)
    dropClearedTeardownTargets blocked while the runner lock was held —
    submit calls this under that lock, so this deadlocks the shard permanently
```

`MarkEntityTornDown` keeps the lock for callers outside it; `markEntityTornDownLocked` serves callers inside. `PrepareSessionPolicy` now takes the lock too, which makes ownership uniform for everything reaching that path and serialises its `TornDownAt` write against a concurrent submit.

## The rest

**Timeout (Claude 1)** — you were right that `teardownVerifyTimeout` was sized for the single post-apply read and then reused for a sequential loop of up to four dial-and-reads. It now scales with the candidate count. I did not switch to a shared client or concurrent checks; scaling the budget is the smaller change and removes the failure mode, and `teardownVerifier` dialling per call is a wider refactor than this release wants.

**Cap ordering (Claude 2)** — flipped to check-then-truncate, as suggested. `onChainTeardownTargets` now returns everything and the caller caps after the filter, so cleared slots cannot consume the budget a live leftover needs.

**Doc (Claude 3)** — corrected. It now says prepare may write `TornDownAt`, that the write is idempotent and only syncs storage to the chain, and that a caller cannot assume prepare touches nothing.

**`deleted` (Claude 4)** — `RevokeSessionGrant` reports it directly now rather than the engine re-deriving it from `hadInstallCall`.

## Verification

- Full offline suite: 25 packages, exit 0.
- Both live replace tests re-run green on Sepolia — they exercise `dropClearedTeardownTargets` on both the prepare and submit paths, including the cleared-candidate case that triggers the deadlock.

chrisli30:
Fixed in `7d2e67ca`. This one was a regression I introduced one commit earlier, and the analysis is exactly right.

Moving the cap out of `onChainTeardownTargets` — so the cleared-entity filter could run before it — left `recheckSupersededGrant` comparing a **capped** prepared set against an **uncapped** storage read. For any wallet holding more than `maxOnChainTeardowns` live candidates that mismatches on every submit, and preparing again reproduces it exactly because nothing about the wallet changed. The grant path closes for that wallet, with an error blaming a race that never happened.

The re-check now applies the same cap in the same stable order.

`TestRecheckToleratesMoreCandidatesThanTheTeardownCap` seeds `cap + 2` candidates and fails without it:

```
--- FAIL: TestRecheckToleratesMoreCandidatesThanTheTeardownCap
    Error: Received unexpected error
    Messages: the uncapped storage set must not read as a change the owner did not sign
```

One clarification on the suggestion to make both sides identical: there is a **second** divergence between prepare and the re-check, and that one is deliberate. Prepare drops entities the chain reports already clear and records `TornDownAt` for them, so they are gone from `onChainTeardownTargets` by the time the re-check reads storage. That is the mechanism the re-check's doc comment already relies on, and it is why the re-check does not need a chain client of its own. Only the cap was out of step.

## Verification

- Full offline suite: exit 0, no failures.
- Both live replace tests re-run green on Sepolia.

PR #738, opened 2026-08-08
docs: point operators at the gateway that actually answers

chrisli30:
`README.md` and `docs/Operator.md` told community operators to dial `aggregator.avaprotocol.org:2206` (and `aggregator-sepolia` for testnet). Neither worked at the time this was opened:

| host | DNS | port 2206 |
| --- | --- | --- |
| `aggregator.avaprotocol.org` | `3.219.223.211` (AWS, per whois `AT-88-Z`) | closed |
| `aggregator-sepolia.avaprotocol.org` | none | — |

Anyone following those instructions got a connection timeout with nothing explaining why.

## The DNS has since been fixed

`aggregator.avaprotocol.org` is **no longer stale** — it is now a CNAME onto the gateway's Railway TCP proxy, applied and verified:

```
$ dig +short CNAME aggregator.avaprotocol.org
zephyr.proxy.rlwy.net.

$ grpcurl -plaintext aggregator.avaprotocol.org:57376 list
aggregator.Node
grpc.reflection.v1.ServerReflection
grpc.reflection.v1alpha.ServerReflection
```

So the address published here answers today. An earlier version of this description named the raw `zephyr.proxy.rlwy.net:57376`, which is what the diff used before the CNAME existed; that is what the branded name resolves to now, and the branded name is preferred because it survives Railway regenerating the proxy host.

## Collapsing the testnet/mainnet split

One gateway serves chains 1, 11155111, 56, 8453 and 84532 and routes per-chain internally, so there is a single operator endpoint rather than one per network.

`docs/Operator.md` states separately that **third-party operator support is mainnet-only**, with testnet coordination handled internally. The endpoint comment there is written accordingly and no longer enumerates the testnet chains the gateway happens to serve, which would have invited exactly the reading that note exists to prevent.

## Dead links removed

`api-explorer.avaprotocol.org` and `api-explorer-sepolia.avaprotocol.org` both fail to connect; the first was an alias onto the retired IP. Telemetry moved with the gateway to `https://api.avaprotocol.org/telemetry` (200), so the two per-chain links collapse into one.

## Route 53, in the same pass

Five dead records removed: four aliases dangling on the deleted `aggregator-holesky` / `aggregator-minato` names — including `api-explorer-minato.avaprotocol.org.avaprotocol.org`, whose doubled suffix meant it could never have resolved — plus `api-explorer`.

## On how long this was broken

I do not know, and an earlier version of this description said "for a while" without evidence. What is verified is the state at the time of checking, above. The only bound is circumstantial: the gateway's TCP proxy was created 2026-05-26. Route 53 keeps no history via its API.

## Note for whoever maintains the gateway service

The published port is Railway-assigned — `TCPProxyCreateInput` has no port argument, and `tcpProxyDelete` is the only related mutation. Deleting that proxy changes the port and breaks these instructions silently. Recorded in `avs-infra/RAILWAY_OPERATIONS.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,9 +82,15 @@ per-chain internally, so there is no longer a separate testnet and mainnet
 address — Ethereum, Sepolia, Base, Base Sepolia and BNB all arrive over the
 same gRPC connection.
 
-The port is not a typo. `aggregator.avaprotocol.org` is a CNAME onto the
-gateway's TCP proxy, and the proxy chooses the public port; DNS cannot carry a
-port, so it has to be written out here.
+The port is not a typo, and it is not ours to choose. The name is a CNAME onto
+the gateway's Railway TCP proxy, which maps a public port Railway assigns onto
+the gateway's own `2206`. Railway's API takes no port argument when a proxy is
+created, so `57376` is what that proxy was given — stable for the life of the
+proxy, and different if one is ever recreated. DNS cannot carry a port, so it
+has to be written out here.
+
+Port `2206` is the gateway's internal listen address and is not reachable from
+outside; operators inside Railway's private network use `gateway.railway.internal:2206`.
 
 The REST API for workflows lives at `https://api.avaprotocol.org/api/v1`, and
 the operator dashboard at
--- a/README.md
+++ b/README.md
@@ -70,19 +70,21 @@ run.
 
 ### Aggregator Address
 
-The aggregator is currently run and managed by the Ava Protocol team. Depend on
-testnet or mainnet, you would need to point your operator to the right address
-in the operator config file.
+The aggregator is run and managed by the Ava Protocol team. Point your operator
+at it in the operator config file:
 
-#### Sepolia Testnet
+```yaml
+aggregator_server_ip_port_address: "zephyr.proxy.rlwy.net:57376"
+```
 
-- aggregator-sepolia.avaprotocol.org:2206
-- [https://api-explorer-sepolia.avaprotocol.org/](https://api-explorer-sepolia.avaprotocol.org/)
+One endpoint serves every chain. The gateway is a single deployment that routes
+per-chain internally, so there is no longer a separate testnet and mainnet
+address — Ethereum, Sepolia, Base, Base Sepolia and BNB all arrive over the
+same gRPC connection.
 
-#### Mainnet
-
-- aggregator.avaprotocol.org:2206
-- [https://api-explorer.avaprotocol.org/](https://api-explorer.avaprotocol.org/)
+The REST API for workflows lives at `https://api.avaprotocol.org/api/v1`, and
+the operator dashboard at
+[https://api.avaprotocol.org/telemetry](https://api.avaprotocol.org/telemetry).
 
 ## Operators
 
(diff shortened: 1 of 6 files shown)


chrisli30:
Updated to the branded endpoint now that the DNS exists.

`aggregator.avaprotocol.org` was still an A record pointing at a retired AWS box (`3.219.223.211`, all ports closed). It is now a CNAME onto the gateway's TCP proxy, so the docs name it directly rather than the Railway-generated host:

```
$ grpcurl -plaintext aggregator.avaprotocol.org:57376 list
aggregator.Node
grpc.reflection.v1.ServerReflection
grpc.reflection.v1alpha.ServerReflection
```

That matters beyond aesthetics: a Railway-generated hostname changes if the proxy is recreated, which is the same drift that left these docs pointing at a dead host to begin with. The CNAME absorbs that. The port still has to be written out — DNS cannot carry one — and the README now says so explicitly so it does not read as a typo.

The same Route 53 pass also removed five dead records in this zone: four aliases dangling on the deleted `aggregator-holesky` / `aggregator-minato` names, and `api-explorer`, which inherited the retired IP and is why that link returned HTTP 000.

chrisli30:
## Correction to this PR's description

I wrote that these hosts had been dead "for a while". **I never measured that**, and it should not have gone in.

What I actually verified, at the time I checked:

- `aggregator-sepolia.avaprotocol.org` — no DNS record
- `aggregator.avaprotocol.org` — resolved to `3.219.223.211` (AWS, per whois `AT-88-Z` — not Hetzner, as I also implied) with ports 80, 443 and 2206 all closed

I do not know when either stopped working. The only bound I have is circumstantial: the gateway's TCP proxy was created **2026-05-26**, so the Railway path has been serving since at least then. Route 53 keeps no history through its API, and CloudTrail would need a live MFA session to check — mine expired.

The fix stands on its own either way: the endpoint the docs named did not answer, and now it does.

## On the port, since it is the obvious question

`57376` is not arbitrary on our side, and it is not selectable. Railway's `TCPProxyCreateInput` accepts exactly three fields:

```
applicationPort   Int (NON_NULL)    <- ours: 2206
environmentId     String (NON_NULL)
serviceId         String (NON_NULL)
```

No port argument, and the only related mutation is `tcpProxyDelete` — there is no update. So the public port is assigned at creation, and recreating a proxy would just draw a different one. `2206` remains the gateway's own listen port (`rpc_bind_address: "0.0.0.0:2206"`), reachable on Railway's private network as `gateway.railway.internal:2206` but not from outside.

The practical consequence, worth knowing before someone tidies up that Networking panel: **deleting the TCP proxy changes the published port and breaks these docs again** — the same drift this PR is fixing.

chrisli30:
All three addressed. Two of them were right about an inconsistency but had the stale side backwards, which is my fault for leaving the description behind the work.

**README.md:77 and docs/Operator.md:38 — `aggregator.avaprotocol.org` is stale, so this will time out.**

It *was* stale when this PR was opened; it is not now. The CNAME was applied and verified before these comments:

```
$ dig +short CNAME aggregator.avaprotocol.org
zephyr.proxy.rlwy.net.

$ grpcurl -plaintext aggregator.avaprotocol.org:57376 list
aggregator.Node
```

The PR description still described the pre-DNS state, which is what these comments read. **Description updated** — the diff was correct. Publishing the branded name rather than `zephyr.proxy.rlwy.net` is deliberate: the Railway hostname changes if the proxy is recreated, which is the same drift that left these docs pointing at a dead host to begin with.

**docs/Operator.md:37 — the comment implies external operators can use testnet, but this doc says mainnet-only.**

Correct, and a contradiction I introduced. Line 47 states third-party operator support is mainnet-only with testnet coordination handled internally, and my comment listed Sepolia and Base Sepolia among the chains the endpoint serves. True of the gateway, misleading on a third-party setup page — it invited exactly the reading that note exists to prevent.

Fixed in `d8a3f1e`: the comment no longer enumerates chains and points at the note instead.

```
# One endpoint, and it is the mainnet one — see the note below on third-party
# operators. The gateway routes per-chain internally, so there is no separate
# address to switch to.
```

PR #749, opened 2026-08-12
feat: map Arbitrum and BNB in Moralis, balances, and token catalog

chrisli30:
Promote Wave A chain maps from staging (#748) onto main.

Cherry-picked 8ff3df89 onto current main instead of opening staging to main directly: staging still has version.go at 4.12.0, which would have downgraded main (4.13.0).

## What's in this PR

- Moralis: 56 to bsc, 42161 to arbitrum; native pricing on both mainnets
- Balance-node chain aliases for Arbitrum One
- ChainIDArbitrumOne + GetNetworkName(42161)
- Local config/worker-arbitrum.example.yaml

Production Railway already has worker-arbitrum connected (v4.13.0). This image change is what makes the new maps live after the next Docker publish.

## Test plan

- [x] Unit tests on #748
- [x] Auth probe chainId 42161 against Railway: JWT aud=42161 (SDK tests/v4/core/auth.test.ts)
- [ ] CI on this PR

### DIFF
--- a/config/worker-arbitrum.example.yaml
+++ b/config/worker-arbitrum.example.yaml
@@ -1,0 +1,35 @@
+# Worker configuration TEMPLATE for Arbitrum One (chain 42161).
+#
+# This is a checked-in template. The real file is gitignored — copy it
+# and fill in the placeholder values:
+#
+#   cp config/worker-arbitrum.example.yaml config/worker-arbitrum.yaml
+#   $EDITOR config/worker-arbitrum.yaml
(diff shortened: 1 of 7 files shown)


chrisli30:
Closing: this was opened against main and skipped the usual feature → staging → main path (and would have risked a version.go downgrade). Review follow-ups are on staging: see the new PR from fix/wave-a-review-followups. After that merges, we cut staging → main.

Issue #676, opened 2026-07-28
feat: per-node retry / error policy for transient failures


Antrikshgwal:
### Summary

Add a per-node **retry / error policy** so a transient failure (RPC blip, HTTP 429/5xx, timeout) retries with backoff instead of failing the whole workflow. Today only infrastructure code retries; user-facing nodes do not.

### Motivation

Ava retries only *infrastructure*: WS RPC reconnect ([engine.go:176](https://github.com/AvaProtocol/EigenLayer-AVS/blob/staging/core/taskengine/engine.go#L176)) and Tenderly simulation ([tenderly_client.go:736](https://github.com/AvaProtocol/EigenLayer-AVS/blob/staging/core/taskengine/tenderly_client.go#L736)). The user-facing REST node even carries a standing `// TODO: Configure retry` ([vm_runner_rest.go](https://github.com/AvaProtocol/EigenLayer-AVS/blob/staging/core/taskengine/vm_runner_rest.go)). A single 429 from an API or a momentary RPC error currently aborts the entire workflow with no recovery.

### Proposed design

**1. `RetryPolicy` message + one shared field on `TaskNode`** (proto, opt-in and additive:
```proto
message RetryPolicy {
  uint32 max_attempts = 1;        // total attempts incl. first (default 1 = current behavior)
  uint32 backoff_ms = 2;          // initial backoff
  double backoff_multiplier = 3;  // exponential factor (default 2.0)
  uint32 max_backoff_ms = 4;      // cap
  repeated string retry_on = 5;   // e.g. ["timeout","http_5xx","http_429","rpc_error"]
}

message TaskNode {
  // ... existing fields 1-3, oneof task_type 10-20 unchanged ...
  RetryPolicy retry_policy = 4;   // field 4 is free; additive, non-breaking
}
```
A single node-level field (not per-node-config) means the policy applies uniformly and there is exactly one place to honor it.

**2. Honored centrally in `executeNode`** ([vm.go:1384](https://github.com/AvaProtocol/EigenLayer-AVS/blob/staging/core/taskengine/vm.go#L1384)), the if/else that dispatches each node to its `runXxx(node)` and returns `(step, err)`. Wrap the retryable `runXxx` calls in an attempt loop; absent policy = 1 attempt = today's behavior, so it is fully backward-compatible.

**`retry_policy` is honored for an explicit allowlist only — it is not globally active.** Although the field lives on `TaskNode`, the wrapper applies it *only* to:

[...]

PR #758, opened 2026-08-14
feat: per-node retry policy for transient read/off-chain failures

Antrikshgwal:
## What

Adds an opt-in per-node `RetryPolicy` that re-runs a node's runner with exponential backoff when it fails with a transient error (RPC blip, HTTP 429/5xx, timeout). Absent policy → a single attempt, identical to today's behavior — fully additive.

```proto
message RetryPolicy {
  uint32 max_attempts = 1;        // total attempts incl. the first (default 1 = no retry)
  uint32 backoff_ms = 2;          // initial backoff before the second attempt
  double backoff_multiplier = 3;  // exponential factor per attempt (default 2.0)
  uint32 max_backoff_ms = 4;      // upper bound on any single backoff
  repeated string retry_on = 5;   // classes: ["timeout","http_5xx","http_429","rpc_error"]
}
```

Attached to a node via `retry_policy` (field 4). Empty `retry_on` enables all classes.

## Safety — idempotent nodes only (#676)

Retry re-invokes the runner, so it is sound only for idempotent reads. The policy is honored for a **read/off-chain allowlist — REST, GraphQL, ContractRead, Balance** — enforced *structurally*: only those four branches in `executeNode` call `executeNodeWithRetry`.

- **`ContractWrite` and `EthTransfer` call their runners directly and are never retried**, so an ambiguous confirmation can't resubmit a UserOp (#676).
- The loop/isolated and direct single-node dispatch paths are intentionally left un-retried (a loop iteration already waits for its own confirmation).

## Error classification

`classifyRetryableError` maps errors to `timeout` / `http_429` / `http_5xx` / `rpc_error`. It's best-effort and string-based — node runners surface upstream failures as wrapped errors rather than typed status codes — and errs toward **not** retrying when unsure. Structured `net.Error` timeouts are matched first.

## Tests

`retry_test.go` (unit, no VM/config, injected sleep → no wall time): classification table, class gating, single-attempt defaults (nil / 0 / 1), retry-then-succeed with backoff assertions, immediate stop on a non-retryable error, exhaustion returns the last failure, and the max-backoff cap.

`go build` + `go vet ./core/taskengine/` clean.

## Notes for review

- No storage-model or key-template changes **in this PR** (model structs unchanged; no keys added/removed by the retry code). `make storage-check` flags pre-existing breaking changes already on the branch (hetznermerge key consolidation, wallet-salt index) — a staging→main release concern, unrelated to this feature.
- `avs.pb.go` diff is `RetryPolicy` codegen + descriptor reflow only — one new struct, nothing removed, no protoc-toolchain version change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -89,6 +89,3 @@ tmp/rehearsal-gateway-backup/
 
 # Local Claude Code agent state (session locks, project-local skills, etc.)
 .claude/
-
-# Editor-local settings
-.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -89,3 +89,6 @@ tmp/rehearsal-gateway-backup/
 
 # Local Claude Code agent state (session locks, project-local skills, etc.)
 .claude/
+
+# Editor-local settings
+.vscode/
--- a/core/taskengine/engine.go
+++ b/core/taskengine/engine.go
@@ -1771,6 +1771,13 @@ func (n *Engine) CreateWorkflow(user *model.User, taskPayload *avsproto.CreateTa
 		return nil, status.Errorf(codes.InvalidArgument, "%s", err.Error())
 	}
 
+	// Constant retry ceilings (limits.go). Retry is a per-execution multiplier on
+	// metered provider calls, so it is bounded like every other one; an over-limit
+	// policy is rejected here rather than silently clamped at execution time.
+	if err := validateRetryPolicies(task.Nodes); err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "%s", err.Error())
+	}
+
 	// Default chain_id to the aggregator's SmartWallet chain when not specified.
 	// In gateway mode the top-level SmartWallet is populated from chains[0]
 	// (mainnet by convention), so a missing chain_id silently routes the task
--- a/core/taskengine/executor.go
+++ b/core/taskengine/executor.go
@@ -266,6 +266,10 @@ func (x *WorkflowExecutor) RunTaskWithContext(ctx context.Context, task *model.W
 	if x.engine != nil {
 		vm.WithChainConfigResolver(x.engine.ResolveSmartWalletConfig)
 	}
+	// Retry backoffs sleep against this context, so a cancelled request aborts a
+	// pending backoff instead of holding the execution open. The async queue path
+	// passes context.Background() and is bounded by the retry ceilings alone.
+	vm.WithContext(ctx)
 
 	// Merge input variables from trigger execution (overrides task-level input variables)
 	if queueData != nil && len(queueData.InputVariables) > 0 {
--- a/core/taskengine/limits.go
+++ b/core/taskengine/limits.go
@@ -101,6 +101,44 @@ const (
 	// quietly processed the first 100 of 5,000 addresses would report success
 	// while skipping most of its job.
 	MaxLoopIterations = 100
+
+	// MaxRetryAttempts bounds a node's RetryPolicy for the same reason
+	// MaxLoopIterations bounds a Loop: retry is a client-controlled multiplier
+	// on metered provider calls *inside* a single execution, which no trigger
+	// floor can reach. Unbounded, `{max_attempts: 100, backoff_ms: 60000}` is
+	// 1h39m of sleeping inside one node.
+	//
+	// 5 attempts is where retrying stops buying anything: with the default 1s
+	// backoff and 2x growth that is 15s of waiting, long enough to ride out a
+	// provider blip or a rate-limit window, and a failure still standing after
+	// four retries is not transient.
+	MaxRetryAttempts = 5
+
+	// DefaultRetryBackoff applies when a policy enables retries without setting
+	// backoff_ms. Without a default, `{max_attempts: 5}` — the most natural
+	// policy to write — retried with *zero* delay, turning one 429 into five
+	// immediate hits.
+	DefaultRetryBackoff = time.Second
+
+	// MaxRetryBackoff caps any single backoff, whether it came from backoff_ms,
+	// max_backoff_ms, or exponential growth. Capping the growing value itself
+	// (not a per-attempt copy) is also what keeps the multiplication from
+	// overflowing time.Duration.
+	MaxRetryBackoff = 30 * time.Second
+
+	// MaxTotalRetryDelay bounds one node's worst-case *total* time asleep, so
+	// the three retry knobs cannot be combined into a long stall that each knob
+	// individually passes. Checked at create time against the exact schedule the
+	// policy produces.
+	MaxTotalRetryDelay = 60 * time.Second
+
+	// MaxExecutionRetryDelay bounds retry sleeping across a whole execution.
+	// Per-node validation cannot see fan-out: a Loop runs its retryable runner
+	// once per element, so MaxLoopIterations nodes each within MaxTotalRetryDelay
+	// still multiply to 100 minutes. Enforced at runtime by retryBudget; once
+	// spent, later nodes fail on their first attempt instead of extending the
+	// execution.
+	MaxExecutionRetryDelay = 5 * time.Minute
 )
 
 // cronSamples is how many consecutive fire times to inspect when deriving a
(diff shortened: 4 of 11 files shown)


chrisli30:
Thanks for this, @Antrikshgwal. The shape is right, and the structural allowlist (only the four read
branches call `executeNodeWithRetry`, so `ContractWrite`/`EthTransfer` cannot reach it)
is the correct way to enforce #676. I verified that by tracing every call site of the four runners, and it holds.

## What needs changing before merge

**1. No ceiling on `max_attempts` / `backoff_ms`, and the sleep can't be cancelled.**
`max_attempts` is an unvalidated `uint32`. `{max_attempts: 100, backoff_ms: 60000}` is
**1h39m of `time.Sleep`** inside `executeNode`, and the async queue path (`Perform` →
`RunTask`) passes `context.Background()` with no deadline, so nothing interrupts it.

This is the big one because it runs against an explicit policy in
`core/taskengine/limits.go` — "Every execution costs money on some metered provider…
Occurrence *is* the cost" — where `MaxLoopIterations = 100` exists precisely to bound work
inside a single execution. Retry adds a new, client-controlled multiplier on metered
provider calls with no bound at all.

Please add constants to `limits.go` (e.g. `MaxRetryAttempts`, `MaxRetryBackoff`, and a
total-retry-delay cap) and validate `retry_policy` at task creation. Following
`MaxLoopIterations`' own reasoning, **reject an over-limit policy rather than silently
clamping it**. Threading a `context.Context` through `executeWithRetry` so cancellation
aborts the sleep would also be good.

**2. `{max_attempts: 5}` on its own retries with *zero* delay.**
`backoff_ms` has no default, unlike `backoff_multiplier` which documents 2.0. Since
`backoff` starts at 0, the `if backoff > 0` guard is never true and no attempt ever
sleeps. We ran it: 5 attempts, 0 of 4 gaps slept. The most natural minimal policy a user
would write turns one 429 into five immediate hits — the opposite of the intent. Default
`backoff_ms` to something sane whenever `max_attempts > 1`.

**3. `strings.Contains(msg, "rpc")` classifies every gRPC status error as retryable.**
`status.Errorf` renders as `"rpc error: code = InvalidArgument desc = …"`, which this
matches. Verified:

```
classifyRetryableError(status.Errorf(codes.InvalidArgument, "task id is malformed"))
  => "rpc_error"
```

So a permanent `InvalidArgument` / `NotFound` / `PermissionDenied` burns the full retry
budget. This codebase returns `status.Errorf` widely, so it isn't a corner case — and it
defeats the "errs toward **not** retrying when unsure" property the doc comment claims.
Suggest `status.FromError()` and allowlisting the genuinely retryable codes
(`Unavailable`, `DeadlineExceeded`, `ResourceExhausted`) instead of substring-matching.

**4. `\b5\d\d\b` matches incidental numbers anywhere in the message.**
Verified:

```
"GET https://api.example.com/v2/tokens?limit=500 returned 404 not found"  => "http_5xx"
```

A genuine 404 gets retried because of a `limit=500` query param. Same class of problem:
`strings.Contains(msg, "429")` isn't word-bounded, and `"eof"` is a bare substring. The
robust fix is to have the REST/GraphQL runners return a typed error carrying the HTTP
status so classification reads a number rather than sniffing prose; short of that, anchor
the match to a status-like context (`status code 5\d\d`, `HTTP 5\d\d`).

**5. Backoff overflows to negative at ~34 attempts and silently stops sleeping.**
Only the local `wait` is capped by `max_backoff_ms` (itself optional) — `backoff` keeps
multiplying unbounded. With `backoff_ms: 1000`, `multiplier: 2.0`, no cap, we measured it
flipping to `-2562047h47m16s` at attempt 34, after which `if backoff > 0` is false and the
remaining attempts become a hot loop. Cap `backoff` itself rather than a local copy.

**6. The safety allowlist has no test.**
The entire #676 argument rests on "only these four branches call the helper" — but nothing
fails if a later refactor wires `ContractWrite` through it. Given the double-spend stakes,
a test asserting that a `ContractWrite`/`EthTransfer` node carrying `max_attempts: 3`
invokes its runner **exactly once** is the highest-value test in the PR, and it's the one
that's missing.

## Discussion — not blocking

**`retry_policy` is silently ignored inside `Loop` nodes.** Loop iterations run through
`executeNodeWithIsolatedVars` / `executeNodeDirect`, neither of which retries — so the
field is accepted and does nothing there. The stated rationale ("a loop iteration already
waits for its own confirmation") describes on-chain writes, not the read nodes on the
allowlist. A REST GET fanned out across a loop is exactly where 429s show up, so as it
stands the feature misses its strongest use case, with no error and no warning. Either
extend it or call it out on the proto field so users aren't surprised.

**Observability.** Only the final attempt's step is returned, so a node that retried four
times over 30s records as one fast success. For an ops feature it'd help to surface the
attempt count and total elapsed on the step.

**`.gitignore`.** The `.vscode/` hunk is unrelated to the feature — and is the sole merge
conflict against `staging`. Dropping it makes the PR conflict-free.

Happy to take another pass once these are addressed!

Antrikshgwal:
898b7e7 covers all six blocking items and the three discussion points. Item 4
also led me to a bug that made the whole feature a no-op for REST; details under
that heading.

## 1. Ceilings and cancellation

I added five constants to `limits.go`, documented next to `MaxLoopIterations`:

| constant | value | bounds |
| --- | --- | --- |
| `MaxRetryAttempts` | 5 | attempts per node |
| `DefaultRetryBackoff` | 1s | applied when `max_attempts > 1` |
| `MaxRetryBackoff` | 30s | any single backoff |
| `MaxTotalRetryDelay` | 60s | one node, worst case |
| `MaxExecutionRetryDelay` | 5m | one whole execution |

`validateRetryPolicies` runs in `CreateTask` next to `validateTriggerFrequency`
and rejects an over-limit policy instead of clamping it, following your point
about `MaxLoopIterations`. The total-delay check walks the schedule the policy
produces, resolving defaults and caps the way `executeWithRetry` resolves them,
so it refuses `{max_attempts: 5, backoff_ms: 20000}` even though each knob sits
under its own ceiling. I left the runtime clamps in place for tasks stored
before the validation existed.

`MaxExecutionRetryDelay` is not on your list. It exists because of the loop
change below: per-node validation cannot see fan-out, so 100 iterations each
within `MaxTotalRetryDelay` still multiply out to 100 minutes. It is a budget
shared across one execution, and once a run spends it, later nodes fail on their
first attempt.

The VM now carries an optional context (`WithContext`, wired in
`RunTaskWithContext`) and backoffs sleep against it. The async queue path still
passes `context.Background()`, so the ceilings are what bound that path.

## 2. Zero-delay retries

`backoff_ms` defaults to 1s when `max_attempts > 1`.
`TestExecuteWithRetry_DefaultsBackoffWhenUnset` asserts that
`{max_attempts: 4}` sleeps 1s, 2s, 4s.

## 3. gRPC status classification

I dropped the substring match on `"rpc"`. Classification reads
`status.FromError` and allowlists three codes: `Unavailable` → `rpc_error`,
`DeadlineExceeded` → `timeout`, `ResourceExhausted` → `http_429`. That last
mapping is the canonical gRPC-to-HTTP one for a rate limit, so a user who turns
`http_429` off also turns off gRPC rate-limit retries. Every other code counts
as a definite answer from the server and skips the prose fallback altogether.
`TestClassifyGRPCStatus` pins nine codes, including
`InvalidArgument`, `NotFound` and `PermissionDenied`.

A status that lost its type upstream and arrives as text still matches, but only
for those same three codes.

## 4. Incidental numbers

Two changes. Classification strips URLs from the message before it looks for a
status, and it reads a status code only in a status-like context (`HTTP 503`,
`status code: 502`). Both of your cases now come back unretryable:

```
GET https://api.example.com/v2/tokens?limit=500 returned status 404 not found  => ""
https://api.example.com/pools/429 returned status 404                          => ""
```

`"eof"` is word-bounded now too.

Your underlying suggestion, read a number rather than sniff prose, is what
turned up the real problem. **The REST runner reports HTTP failures as
`(step, nil)` with `success = false`.** `err` is nil, so no HTTP failure ever
reached the retry loop: `http_429` and `http_5xx` were dead classes for the node
they matter most for. Classification now takes the step alongside the error and
reads the status code out of the REST response envelope
(`{status, statusText, url, headers, data}`), so that path compares a number
rather than a regex. `TestRetryClassFor_FailedStepWithoutError` covers it,
including a 404 whose URL carries `limit=500`.

## 5. Backoff overflow

I capped `backoff` itself rather than a per-attempt copy, and `nextBackoff`
bounds-checks in float64 before it converts, so the value saturates instead of
wrapping. `TestNextBackoff_SaturatesInsteadOfOverflowing` runs 200 growth steps
and asserts it never goes non-positive or past the cap. Capping `max_attempts`
at 5 puts the 34-attempt case out of reach, but the growth function holds on its
own.

## 6. Allowlist test

Two tests, plus a guard you did not ask for.

`TestRetryAllowlist_WriteNodesRunExactlyOnce` parses `vm.go`, attributes each
`executeNodeWithRetry` call site to the node-type branch containing it, and
fails if one lands outside {`GetRestApi`, `GetGraphqlQuery`, `GetContractRead`,
`GetBalance`} or outside any node-type branch. It also fails if a read branch
loses its retry wiring. `TestRetryAllowlist_WriteRunnersNeverWrapped` asserts
that `runContractWrite` and `runEthTransfer` never reach the helper. I checked
both by wiring `ContractWrite` through it:

```
retry_allowlist_test.go:100: vm.go:4973:30: node branch GetContractWrite() routes its
  runner through executeNodeWithRetry, but only idempotent reads may be retried
retry_allowlist_test.go:143: vm.go:4973:30: runContractWrite is passed to
  executeNodeWithRetry; on-chain writes must never be re-invoked (#676)
```

This reads source rather than counting runner invocations at runtime, which is a
trade I want to flag. Attempts are invisible in the execution logs, since retry
returns only the final step, and driving a real `ContractWrite` to a *retryable*
failure needs config and a socket. The AST test checks the claim the safety
argument rests on, "only these branches call the helper", across all three
dispatchers, including the two loop paths the first version left out.

The extra guard: a `retry_policy` on a node that cannot honor it now fails at
creation, so nobody can store a write node carrying a policy that looks live.

## Discussion points

**Loops.** Extended rather than documented away. The Loop's policy propagates to
the synthetic per-iteration node for read runners, so a REST GET fanned out
across a loop retries while `ContractWrite` and `EthTransfer` loops still do
not. `MaxExecutionRetryDelay` keeps that fan-out bounded.

**Observability.** A retried step rewinds `start_at` to the first attempt, so
`end_at - start_at` reports true elapsed time, and its log carries
`Retry: N attempts over X`. I kept this off the proto. Tell me if you would
rather have structured fields you can query.

**`.gitignore`.** Dropped. The file matches `staging` byte for byte again, so
the PR carries no conflict.

## Verification

`go build ./...` and `go vet ./...` come back clean, and the retry, validation
and allowlist tests pass. Four `TestLoopNode_ContractWrite*` / `EthTransfer`
tests fail in my environment on `testConfig is nil`; I stashed the branch and
saw the same four fail on the baseline, so they predate this work. Nothing here
touches storage keys or model structs.

chrisli30:
898b7e7 clears all six. I verified each against current `staging` rather than reading the diff.

**First, the CI.** The 26 red checks are not your code — every secret arrives blank on a
fork PR (`ALCHEMY_API_KEY: `), so `setup-test-config` exits 1 and no Go test ever runs.
Ignore it; we run the suite on our side.

## Verified fixed

1. **Ceilings + cancellation** — constants in `limits.go`, rejected at create time, and
   backoffs now sleep against a cancellable context.
2. **Zero-delay retries** — `backoff_ms` defaults to 1s.
3. **gRPC classification** — `status.Errorf(codes.InvalidArgument, …)` → `""`. The
   three-code allowlist is the right call.
4. **Incidental numbers** — both my cases plus seven adversarial ones I added all come
   back correct.
5. **Backoff overflow** — saturates instead of wrapping.
6. **Allowlist test** — I mutation-tested it at a site you didn't use (`EthTransfer` in
   `executeNode`) and both tests fired.

Merged onto current `staging`: build and vet clean, all 19 retry tests pass under `-race`,
and the full `core/taskengine -short` run has an identical failure set to baseline — the
failures you saw are pre-existing, as you said.

**The REST bug is a genuinely good find**, and better than the item that led to it.
`vm_runner_rest.go:848` returns `(step, nil)` on HTTP ≥400, so `http_429` and `http_5xx`
were dead for the node they matter most for. GraphQL, ContractRead and Balance all return
real errors, so scoping the fix to the REST envelope is right.

Nice work! Putting the ceilings in `limits.go` with a worst-case schedule check, rather
than ad-hoc clamps, is exactly right. This is good to merge now.

PR #760, opened 2026-08-20
feat: chain-scoped wallet listing and grant chain validation

chrisli30:
Three commits on the wallet/grant chain-routing path. Together they complete the gateway half of Studio Phase 4: a JWT minted for one chain can prepare a grant on another, and now list that chain's wallets, without a second signature.

## `feat(wallets)` — list wallets per chain via `?chainId`

Wallet records are chain-scoped, but `GET /wallets` always read the gateway default: the engine hardcoded `defaultChainID()` behind a TODO. A wallet created on one chain was invisible unless that chain happened to be the gateway's, so a client holding a token for one chain had to mint a second one purely to read a list.

Adds `ChainIdQuery` to `GET /wallets` on the precedence `CreateWallet` already uses — explicit value, then JWT `aud`, then gateway default — and makes the engine follow the resolved chain.

`ListWallets` now takes `*model.User` rather than a bare owner address, since that is what carries the chain. All three internal callers already had the user in scope and used `user.Address` as the owner, so they gain the caller's chain instead of always reading the gateway's.

Everything chain-dependent moves to the resolved chain's config: the factory the default wallet is derived against, the per-owner wallet cap, and the storage prefix. Derivation prefers the per-chain `ChainStateReader` and falls back to the package-level `rpcConn` only for the default chain it belongs to. With neither, the default wallet is omitted rather than derived against the wrong chain's factory — a wrong address there would be indistinguishable from a real one in the response.

## `fix(policies)` — refuse grants on chains the gateway does not serve

Prepare and submit took `chainId` straight from the request body with no validation, so a grant could be allocated, signed by the owner, and stored for a chain this gateway has no config for.

Nothing could ever send under it. `ResolveSmartWalletConfig` silently returns the **default** chain's config for an unknown chain, while the EIP-712 domain commits to the real `chainId` — the MA v2 wallet address is identical across chains, so `chainId` is the only field separating a Base grant from a Sepolia one. The signature is sound; there is simply no bundler or RPC behind it. The grant also lands in `sp:<chainID>:*`, which no read path here comes back for.

`resolveUserChainID` already validates the chain from the **signed** JWT `aud` against `knownChainIDs()`, citing orphaned buckets and unbounded keyspace growth. The request body is a weaker input and had no check at all.

**Refused rather than resolved to the default chain**, which is what `resolveUserChainID` does for a wallet read. Falling back is the wrong remedy for a grant — worse than the bug: the owner asks for one chain, signs a prompt naming it, and is handed spending authority on another. Refusing at prepare means no signature is collected for a grant that cannot work. Submit is gated independently: it re-derives from the client's echo and cannot inherit prepare's verdict.

Surfaces as `400 POLICIES_CHAIN_NOT_SERVED`.

## `test(rest)` — prepare uses body chainId, not JWT aud

Studio R4b. The grant chain is base-sepolia while both the JWT `aud` and the gateway default are the test chain, so the assertion excludes every wrong source rather than just the `aud` — written the weaker way, a handler that ignored the body entirely passed.

## Verification

- `make test/unit` — full suite, zero failures
- `make storage-check` vs `origin/main` — key templates and persisted model structs unchanged
- Mutation-checked throughout: pinning prepare to the JWT aud or the gateway default, dropping the `?chainId` override, pinning the engine back to `defaultChainID()`, removing the submit gate, or degrading the chain gate to a silent fallback each fail the new tests

## Release + rollout

Prefix is `feat:` per the highest-impact rule across the bundle (was `fix:` before the ListWallets work landed on `staging`).

Two API-contract notes, neither treated as breaking:
- The policy gate turns some previously-`200` prepare/submit calls into `400`. The requests it now rejects produced grants that were unusable by construction, so no functioning integration could have depended on them.
- `GET /wallets` returns the JWT `aud` chain's wallets where it previously returned the gateway default's. For any single-chain deployment, and for any caller whose `aud` is the gateway default, that is the same set.

**Downstream ordering:** Studio [#1578](https://github.com/AvaProtocol/studio/pull/1578) must not merge until this is *deployed*, not merely merged — `main` cuts a pre-release, `avs-infra` promotes. #1578 flips `LIST_WALLETS_ACCEPTS_CHAIN_ID`, skipping the mandate SIWE; against a gateway that still ignores `?chainId` the grant path would proceed against the default chain's wallets with no SIWE in front of it.

### DIFF
--- a/aggregator/rest/generated/server.gen.go
+++ b/aggregator/rest/generated/server.gen.go
@@ -67,7 +67,7 @@ type ServerInterface interface {
 	RunTrigger(ctx echo.Context) error
 	// List the authenticated user's smart wallets
 	// (GET /wallets)
-	ListWallets(ctx echo.Context) error
+	ListWallets(ctx echo.Context, params ListWalletsParams) error
 	// Derive (and persist) a smart wallet address from (owner, salt, factory)
 	// (POST /wallets)
 	CreateWallet(ctx echo.Context) error
@@ -515,8 +515,17 @@ func (w *ServerInterfaceWrapper) ListWallets(ctx echo.Context) error {
 
 	ctx.Set(PartnerAssertionScopes, []string{})
 
+	// Parameter object where we will unmarshal all parameters from the context
+	var params ListWalletsParams
+	// ------------- Optional query parameter "chainId" -------------
+
+	err = runtime.BindQueryParameter("form", true, false, "chainId", ctx.QueryParams(), &params.ChainId)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("Invalid format for parameter chainId: %s", err))
+	}
+
 	// Invoke the callback with all the unmarshaled arguments
-	err = w.Handler.ListWallets(ctx)
+	err = w.Handler.ListWallets(ctx, params)
 	return err
 }
 
@@ -1641,6 +1650,7 @@ func (response RunTrigger401ApplicationProblemPlusJSONResponse) VisitRunTriggerR
 }
 
 type ListWalletsRequestObject struct {
+	Params ListWalletsParams
 }
 
 type ListWalletsResponseObject interface {
@@ -3193,8 +3203,10 @@ func (sh *strictHandler) RunTrigger(ctx echo.Context) error {
 }
 
 // ListWallets operation middleware
-func (sh *strictHandler) ListWallets(ctx echo.Context) error {
+func (sh *strictHandler) ListWallets(ctx echo.Context, params ListWalletsParams) error {
 	var request ListWalletsRequestObject
+
+	request.Params = params
 
 	handler := func(ctx echo.Context, request interface{}) (interface{}, error) {
 		return sh.ssi.ListWallets(ctx.Request().Context(), request.(ListWalletsRequestObject))
--- a/aggregator/rest/generated/types.gen.go
+++ b/aggregator/rest/generated/types.gen.go
@@ -1900,6 +1900,13 @@ type GetTokenParams struct {
 	ChainId *ChainIdQuery `form:"chainId,omitempty" json:"chainId,omitempty"`
 }
 
+// ListWalletsParams defines parameters for ListWallets.
+type ListWalletsParams struct {
+	// ChainId The chain to operate on (a single value). Omit to use the aggregator
+	// default (the request's JWT `aud` chain, then the gateway default).
+	ChainId *ChainIdQuery `form:"chainId,omitempty" json:"chainId,omitempty"`
+}
+
 // ListWalletPoliciesParams defines parameters for ListWalletPolicies.
 type ListWalletPoliciesParams struct {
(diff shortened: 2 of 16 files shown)


chrisli30:
Addressed both minor observations from the review with one change (e6431f7f).

The served-chain list was the root of both: it made the refusal a config disclosure to any authenticated caller (since this check deliberately runs ahead of the ownership lookup), and it was nondeterministically ordered because `knownChainIDs()` iterates a map.

The actionable half of the error is the chain the caller named — that stays. The rest moves to a `Warn` log, so an operator debugging a refusal still has the served set. With nothing disclosed, the cheap-check-first ordering stands on its own merits.

Left the long doc comment as-is, per the review's own reasoning.

chrisli30:
Good catch — accepted all three, fixed in 73192c20.

Both paths resolve an explicit chain and then didn't use it for the runner lookup: `simChainID` (what `NewVMWithData` gets its smart-wallet config from) in `SimulateWorkflowWithContext`, and `settings.chain_id` in `runProcessingNodeWithInputs`. A runner registered only on the requested chain read as unknown — simulate fell through to derivation and simulated as a different sender, RunNodeImmediately rejected it.

One correction to the framing: this isn't introduced by the `?chainId` work. Before it, both sites called `ListWallets` with a bare owner and the engine listed `defaultChainID()` unconditionally, so they were wrong for *any* non-default-chain request. Passing the user narrowed that to the audience chain; passing the effective chain closes it. The suppressed comment on line 3862 was the same bug and is fixed by the same `simUser`.

`settingsChainID` is hoisted out of the config-resolution `if` so the lookup uses the identical value the VM did, rather than re-deriving it.

No dedicated test: exercising it needs an AA node simulated cross-chain with a runner registered only on the target chain, which is integration-scale. Existing simulate/run-node coverage passes, full `make test/unit` clean.

chrisli30:
Added the dedicated coverage in 935ea576 — `core/taskengine/simulate_runner_chain_test.go`, one test per path. Both register the runner **only** on the non-audience chain so the two cannot agree by accident, and both fail against the pre-fix code (verified by reverting each routing change in turn).

Writing them turned up something worth recording: **the two paths fail differently, and the simulate one fails silently.**

A runner missing from the listed chain does not stop simulate's resolution. It falls through to "first wallet owned by the user", which on a chain with working derivation is the owner's *default* wallet — so a cross-chain simulation ran as the wrong sender and reported success rather than reporting an unknown runner. My first attempt asserted on an error and failed for that reason: there is no error to assert on.

So the observable there is *which sender was chosen*. The workflow carries a contractWrite node to make `taskNodesRequireAASender` true, but only executes a customCode node that echoes `aa_sender` — the gate's verdict becomes readable without simulating a real write, and the test stays out of `//go:build integration`.

`RunNodeImmediately` is loud by contrast: a missing runner reaches the salt-0..4 derivation scan and is rejected outright, so that test asserts on the rejection.

Worth noting the silent variant was the more damaging of the two, and neither bot review caught that distinction — the inline comments described both as the runner being "missed" or "rejected".

`make test/unit` clean; full `core/taskengine` package green.

PR #765, opened 2026-08-22
fix: stop reusing session-grant entities that still hold on-chain hooks

### LINKED ISSUE #763: fix: reused validation entity inherits stale on-chain hooks, so storage and chain disagree about a grant
Session grant entity allocation is storage-derived, so a reused validation entity can inherit the account's stale hooks. Storage then says the grant is valid while the chain enforces the previous grant's time range, and every UserOp is rejected as expired with nothing pointing at the cause.

Observed on the Sepolia e2e fixture, gateway `4.17.1`, runner `0x209eb31c199bEB4c386eF83CF442DE1a00667a1F` (salt 0, factory `0x0000…6fecd`, owner `0x804e…1557`).

## Symptom

Storage and chain disagreed about entity 2:

| | validUntil |
|---|---|
| Stored `SessionPolicy` (status `active`) | 2027-08-21 |
| Account's installed `TimeRangeValidationHook` | **2026-08-13** |

Every UserOp signed for that entity was rejected:

```
eth_sendUserOperation: validation simulation failed:
User Operation expired or has an invalid time range.
validUntil: 2026-08-13 20:04:33 UTC, validAfter: 1970-01-01 00:00:00 UTC
```

Sentry `EIGENLAYER-AVS-2C`. This is not the `SESSION_POLICY_EXPIRED` case fixed in `cc19bec4` — storage believed the grant was live, so no storage-side check could catch it. The wallet was funded (0.215 ETH, 56.5 USDC), so the suite's "fund this address" hint was a red herring.

Confirmed by experiment, in order:

1. Re-ran the UserOp tests with the 2027 grant in place → same `2026-08-13` rejection, so not a within-run race.
2. Revoked it and re-granted, which moved the fixture **entity 2 → entity 3**.
3. Re-ran → **20/20 pass**.

Moving entity is what fixed it. Nothing about the grant's contents changed.

## Mechanism

`NextSessionEntityID` (`core/taskengine/session_policy.go`) derives the next entity purely from stored records:

```go
next := uint32(1)
for _, p := range policies {
    if p.Runner == nil || *p.Runner != wallet { continue }
    if p.EntityID >= next { next = p.EntityID + 1 }
}
```

Its own comment states the invariant that makes this safe:

> Revoked policies still count: their entity may hold module state that was never cleaned up, and reusing it would collide with the leftovers.

`supersedeUsablePolicies` (session_policy.go:178-188) is careful to preserve it — it revokes rather than deletes, precisely so an entity stays spoken for.

But `RevokeSessionGrant` (`core/taskengine/session_grant.go:465`) **deletes** the record on one path:

```go
// Pending without InstallCall (should not occur after submit): deleted.
return true, false, db.Delete(key)
```

A deleted record is invisible to `NextSessionEntityID`, so its entity is handed out again. If the chain still has that entity installed — which it will whenever the earlier install landed but its record went away — the new grant is installed "into" an occupied entity and the account keeps the OLD hooks. The gateway marks the grant `active` and starts signing for it.

The asymmetry is the core of it: **teardown is verified, allocation is not.**

- `VerifySupersededTeardown` → `TeardownVerifier` → `aa.EntitySignerOnChain` already reads back whether a *superseded* entity is really gone, and deliberately treats an unknown result as not-cleared.
- Nothing performs the mirror check when *allocating*: is the entity we are about to install into actually free on chain?

## Decision

**A + B, then a bounded send/apply check.** These are not alternatives — this
issue originally listed three options as if picking one would do, and that is
wrong. A and B close *allocation*, prospectively. Neither helps a grant that is
**already** stored wrong, which is exactly the state this was filed from: entity
2 was active-and-2027 in storage before anything went looking. Only a chain read
catches that one.

### B. Stop deleting (do this regardless)

`supersedeUsablePolicies` already refuses to delete so entities stay reserved;
`RevokeSessionGrant`'s pending-without-InstallCall path breaks that same
invariant. Make it revoke instead. Trivial, matches the documented intent, and
covers the one path we can name.

Not sufficient alone: a wiped DB, a botched migration, or an install that never
got a record still leaves an occupied entity for `NextSessionEntityID` to reuse.

### A. Verify at allocation

On grant prepare, read the candidate entity with the helper that already exists —
`aa.EntitySignerOnChain` — and treat a non-zero signer as occupied. Same
fail-closed rule `VerifySupersededTeardown` uses: **a failed read counts as
occupied**, never as free.

If the candidate is occupied, bump to the next id and retry rather than failing —
nobody cares that entity 2 was leftover. **Bound that loop.** `NextSessionEntityID`
returns `max(stored)+1` and every bump costs an RPC, so an account with several
unstored leftovers would turn prepare into an unbounded read loop. Try a fixed
number of candidates, then fail with a typed error.

The cost lands in the right place: one read on a rare operation, and the doomed
grant is never signed. Same shape as `POLICIES_CHAIN_NOT_SERVED` and #761 —
refuse before the user commits, not at the bundler.

### C. Send/apply check — dependent follow-up, not a co-equal option

A+B would NOT have saved the entity-2 grant, so AC3 needs this. It is gated on a
reader that **does not exist yet**:

- `aa.EntitySignerOnChain` reads `signers(uint32,address)` on the
  **SingleSignerValidationModule**. That answers "is this entity occupied", which
  is all A needs.
- For the installed *window* there is no read at all — `core/chainio/aa` has only
  `PackTimeRangeInstallData` / `PackTimeRangeUninstallData`, both writes.

The read is thin and mirrors the existing helper. TimeRangeModule
(`0x00000000000082B8e2012be914dFA4f62A0573eA`) exposes a public mapping:

```solidity
struct TimeRange { uint48 validUntil; uint48 validAfter; }
mapping(uint32 entityId => mapping(address account => TimeRange)) public timeRanges;
```

so the generated getter is `timeRanges(uint32,address)` returning
`(uint48 validUntil, uint48 validAfter)` — the same call shape as
`signers(uint32,address)`.

With that: on first use after install (or wherever the grant is marked active),
compare the installed window against the stored `ValidUntil`; on mismatch raise a
typed error naming the entity and BOTH windows instead of letting Alchemy's
expiry string through. Cache that a grant has been chain-checked so this is not
paid per UserOp. Promote it to every send only if drift turns up from writers we
do not control.

**The reader is needed regardless of C**: the existing production population is
already drifted and A+B do not heal it. Finding which stored grants disagree with
their account requires exactly this call, so build it once and use it for both
the check and a one-time sweep.

### Sequence

1. **B** — hygiene, trivial, no new reads.
2. **A** — closes allocation, bounded bump loop.
3. **`EntityTimeRangeOnChain`** — the missing reader.
4. **C + sweep** — send/apply mismatch check, plus a one-off audit of existing records.

### Rejected

- **C alone.** It is the check that would have typed-errored this failure, but as
  the primary fix it is the wrong layer: a chain read on every trigger and
  withdraw to catch a rare bug; a colliding grant still gets signed and stored, so
  every send fails until someone re-grants — the loop #761 exists to stop; and
  `NextSessionEntityID`'s documented invariant stays broken.
- **Folding expiry into `Usable()`.** Rejected in `cc19bec4`; an expired record
  still has to keep its entity reserved. `TestUsableIgnoresExpiryOnPurpose` pins it.
- **Closing this on B alone** because the delete path is the one we can name.

## Acceptance

- [ ] **(A)** A grant can never be allocated an entity the account already has installed. A failed chain read counts as occupied, and the retry loop is bounded.
- [ ] **(B)** Deleting a session policy record cannot silently free an entity that is still installed on chain.
- [ ] **(C)** Regression test: a stored grant whose `ValidUntil` disagrees with the entity's installed hook does not reach the bundler. A+B cannot satisfy this for already-stored records.
- [ ] **(sweep)** Existing stored grants whose installed window disagrees with storage are found and re-granted onto clean entities.
- [ ] Live Sepolia check on a funded MA v2 runner: grant, revoke without on-chain cleanup, grant again, and confirm the second grant does not inherit the first's time range.

## Not in scope

- `SESSION_POLICY_EXPIRED` (fixed in `cc19bec4`) — that is the case where *storage itself* knows the grant is expired. This issue is the case where storage does not.
- The ava-sdk-js fixture scripts (`scripts/inspect-fixture-grant.ts`, `scripts/refresh-fixture-grant.ts`). Those move the fixture to a clean entity and are a **workaround**, not the fix.
- Production wallet-cap 429s in the e2e suite (leftover `nextTestSalt()` on isolated EOAs) — unrelated.

## Evidence pointers

| Path | File |
|---|---|
| Entity allocation from storage only | `core/taskengine/session_policy.go` `NextSessionEntityID` |
| Invariant that entities stay reserved | `core/taskengine/session_policy.go:178-188` `supersedeUsablePolicies` |
| The delete that frees an entity | `core/taskengine/session_grant.go:465` `RevokeSessionGrant` |
| Teardown IS verified (the missing mirror) | `core/taskengine/session_grant.go:639` `VerifySupersededTeardown` |
| The chain read to reuse | `core/chainio/aa` `EntitySignerOnChain` |

chrisli30:
Closes #763.

## Why

`NextSessionEntityID` derived the next validation entity from stored records only, and `RevokeSessionGrant` deleted pending-without-InstallCall rows. Installing a new grant into an entity the account still held kept the **old** TimeRangeModule window.

Observed on the Sepolia e2e fixture: storage said entity 2 was valid until 2027-08-21; the account enforced 2026-08-13. Every UserOp failed as "User Operation expired or has an invalid time range" with nothing pointing at the drifted record. Sentry `EIGENLAYER-AVS-2C`. This is not the storage-known expiry #764 already refuses.

A and B are not substitutes. They close **allocation**, prospectively. Neither heals a grant that is already stored wrong — which is the state this was filed from. That needs a chain read (C) and a sweep.

## What

Sequenced **B → A → reader → C + sweep**:

- **B.** `RevokeSessionGrant` never deletes. Revoked records stay so `NextSessionEntityID` keeps the entity reserved. Incomplete revoked rows are allowed by `Validate()`.
- **A.** Prepare/submit allocate with `NextFreeSessionEntityID`: storage max+1, then a bounded on-chain occupancy probe (signer **or** deferred nonce sequence, fail-closed on read error, cap 8, then `SESSION_POLICY_ENTITY_OCCUPIED`). Wired on the Engine only after `InstallSessionResolver()`, so unit tests stay offline.
- **Reader.** `aa.EntityTimeRangeOnChain` reads `TimeRangeModule.timeRanges(uint32,address) → (uint48, uint48)`.
- **C.** For applied grants not yet cached, the send path compares the installed window to stored `ValidUntil`. Mismatch is `SESSION_POLICY_CHAIN_WINDOW_MISMATCH` (never reaches the bundler). Match sets `Grant.ChainWindowChecked`.
- **Sweep.** `FindDriftedSessionGrants` plus `scripts/sweep_session_windows` list already-stored drift. Re-grant is the owner's — this process cannot produce the signature.

`Usable()` stays status-only. `TestUsableIgnoresExpiryOnPurpose` still pins that.

See `docs/changes/20260821-session-entity-reuse-drift.md`.

## Test plan

- [x] Occupancy: skip occupied, fail-closed on read error, probe cap → typed error
- [x] Revoke without delete keeps the entity reserved
- [x] Send path refuses a stored/chain window mismatch; matching window is cached
- [x] Sweep reports mismatch and ignores matching windows
- [x] `EntityTimeRangeOnChain` encoding + short-return errors
- [x] `SESSION_POLICY_CHAIN_WINDOW_MISMATCH` is classified client-fixable
- [x] `make storage-check` vs staging: additive only (`ChainWindowChecked omitempty`, scan prefix `sp:%d:`)
- [ ] Live Sepolia: grant, revoke without on-chain cleanup, grant again — second grant must take a new entity and not inherit the first's time range

## Out of scope

- `SESSION_POLICY_EXPIRED` (shipped in #764) — storage itself knows the grant is expired
- Auto re-grant of drifted records (needs the owner signature)
- The ava-sdk-js fixture refresh scripts (workaround, not the fix)

### DIFF
--- a/aggregator/rest/handlers_policies.go
+++ b/aggregator/rest/handlers_policies.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"math/big"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -411,6 +412,11 @@ func mapPolicyError(err error) error {
 				" List this wallet's policies and revoke the older usable ones, or grant again to replace them."}
 	case errors.Is(err, taskengine.ErrGrantSignerMismatch):
 		return badRequest("POLICIES_BAD_SIGNATURE", "Signature does not match", err.Error())
+	case strings.Contains(err.Error(), taskengine.SessionPolicyEntityOccupiedCode):
+		// Bounded occupancy probe exhausted (#763 A). The account still has
+		// leftover entities; granting cannot allocate a free one. 400 so
+		// the owner is not asked to sign a doomed payload.
+		return badRequest(taskengine.SessionPolicyEntityOccupiedCode, "Validation entity occupied", err.Error())
 	default:
 		return badRequest("POLICIES_REJECTED", "Policy request rejected", err.Error())
 	}
--- a/core/chainio/aa/ma_v2_entity_state.go
+++ b/core/chainio/aa/ma_v2_entity_state.go
@@ -52,6 +52,36 @@ func EntitySignerOnChain(ctx context.Context, client ContractCaller, account com
 	return common.BytesToAddress(out[12:32]), nil
 }
 
+// EntityTimeRangeOnChain returns the TimeRangeModule window for
+// (entity, account). validUntil / validAfter are unix seconds (uint48 on
+// chain, ABI-encoded as two 32-byte words). Both zero means the entity
+// has no time-range hook installed.
+//
+// A read failure is returned rather than folded into zeros: zeros mean
+// "no window", and a caller acting on a transient RPC error would treat
+// a live grant as unwindowed (#763).
+func EntityTimeRangeOnChain(ctx context.Context, client ContractCaller, account common.Address, entity uint32) (validUntilSec, validAfterSec uint64, err error) {
+	if client == nil {
+		return 0, 0, fmt.Errorf("no chain client")
+	}
+	data := crypto.Keccak256([]byte("timeRanges(uint32,address)"))[:4]
+	data = append(data, common.LeftPadBytes([]byte{
+		byte(entity >> 24), byte(entity >> 16), byte(entity >> 8), byte(entity),
+	}, 32)...)
+	data = append(data, common.LeftPadBytes(account.Bytes(), 32)...)
+
+	module := TimeRangeModuleAddress()
+	out, err := client.CallContract(ctx, ethereum.CallMsg{To: &module, Data: data}, nil)
+	if err != nil {
+		return 0, 0, fmt.Errorf("reading time range of entity %d on %s: %w", entity, account.Hex(), err)
+	}
+	if len(out) < 64 {
+		return 0, 0, fmt.Errorf("timeRanges(%d, %s) returned %d bytes, want at least 64",
+			entity, account.Hex(), len(out))
+	}
+	return new(big.Int).SetBytes(out[:32]).Uint64(), new(big.Int).SetBytes(out[32:64]).Uint64(), nil
+}
+
 // EntryPointNonce returns the EntryPoint's current nonce for (account, key) —
 // the full 256-bit value, whose low 64 bits are the sequence.
 //
--- a/core/chainio/aa/ma_v2_entity_state_test.go
+++ b/core/chainio/aa/ma_v2_entity_state_test.go
@@ -116,3 +116,33 @@ func TestEntityDeferredNonceSequenceReportsReadFailures(t *testing.T) {
 		common.HexToAddress("0x0643f9d156e2CE584825511C548649F3289619c3"), 1)
 	require.Error(t, err, "no client is not evidence that an entity is free")
 }
+
+func TestEntityTimeRangeOnChainDecodesBothWords(t *testing.T) {
+	account := common.HexToAddress("0x209eb31c199bEB4c386eF83CF442DE1a00667a1F")
+	const entity = uint32(2)
+	until := uint64(1785541473) // 2026-08-13-ish
+	after := uint64(0)
+
+	result := append(word(new(big.Int).SetUint64(until)), word(new(big.Int).SetUint64(after))...)
+	caller := &recordingCaller{result: result}
+
+	gotUntil, gotAfter, err := EntityTimeRangeOnChain(context.Background(), caller, account, entity)
+	require.NoError(t, err)
+	require.Equal(t, until, gotUntil)
+	require.Equal(t, after, gotAfter)
+
+	require.Equal(t, TimeRangeModuleAddress(), caller.to)
+	require.Len(t, caller.data, 4+32+32, "timeRanges(uint32,address)")
+	require.Equal(t, account, common.BytesToAddress(caller.data[4+32+12:4+32+32]))
+	require.Equal(t, big.NewInt(int64(entity)), new(big.Int).SetBytes(caller.data[4:4+32]))
+}
+
+func TestEntityTimeRangeOnChainReportsReadFailures(t *testing.T) {
+	account := common.HexToAddress("0x209eb31c199bEB4c386eF83CF442DE1a00667a1F")
+	caller := &recordingCaller{result: word(big.NewInt(1))} // 32 bytes, need 64
+	_, _, err := EntityTimeRangeOnChain(context.Background(), caller, account, 1)
+	require.Error(t, err)
+
+	_, _, err = EntityTimeRangeOnChain(context.Background(), nil, account, 1)
+	require.Error(t, err, "no client is not evidence that an entity has no window")
+}
--- a/core/taskengine/engine.go
+++ b/core/taskengine/engine.go
@@ -268,6 +268,11 @@ type Engine struct {
 	assignmentMutex      *sync.RWMutex     // protects task assignments
 
 	smartWalletConfig *config.SmartWalletConfig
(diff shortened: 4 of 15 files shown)


chrisli30:
Blocked on #766 — found while reviewing this branch against live chain state.

The window check here (C) refuses a grant whose stored `ValidUntil` disagrees with the installed window. The Sepolia fixture's entity 3 is exactly that state: signer and allowlist hooks on chain, no TimeRange hook.

The problem is that the same state also makes teardown impossible. `BuildOnChainRevokeCleanup` derives `hookUninstallData` from the **stored** install call, so it emits one entry per hook the grant intended (3) while the account has 2 — `uninstallValidation` reverts with `ArrayLengthMismatch()` (`0xa24a13a6`). Since teardown is part of every later grant's install/replace batch, that batch reverts too:

```
SESSION_GRANT_INSTALL_FAILED: deferred grant install/replace did not land
(new entity not installed; prior entities not torn down): ... AA23 reverted
```

So on a runner in this state, this PR refuses the grant **and** the remedy it recommends ("grant again onto a free entity") cannot execute. That turns a silent wrong-window into a hard stop with no recovery path.

Nothing wrong with the code here — #766 is a separate defect this branch surfaces rather than causes. But I would land #766 first, or at minimum know the count from the production sweep before deploying C.

Also note `SESSION_GRANT_INSTALL_FAILED` was only visible in Railway logs; the API returned a bare `{"status":"failed"}` until staging `14043953` fixed the adapter dropping the message.

Issue #766, opened 2026-08-22
fix: revoke cleanup is derived from stored hooks, so a partially-installed entity can never be torn down


chrisli30:
A revoke's on-chain cleanup payload is derived from the grant's **stored** `InstallCall`, so it lists one teardown entry per hook the grant *believes* it installed. When the account actually has fewer, `uninstallValidation` reverts with `ArrayLengthMismatch()` and the entity can never be cleared.

That is not a cosmetic failure. Teardown is part of every later grant's install/replace batch, so once one entity is in this state the runner accepts **no further grants at all**.

Found while investigating why the Sepolia e2e fixture stopped accepting grants after #763's diagnosis. Gateway `4.17.2`, runner `0x209eb31c199bEB4c386eF83CF442DE1a00667a1F`, owner `0x804e…1557`.

## The loop

1. **A partial install.** Entity 3 has its signer and allowlist hooks on chain but **no TimeRange hook**. Storage records the grant as applied with all three.

   ```
   entity 1  window=2026-08-06T10:57:21Z  signer=0x82F2Dd9a…   complete
   entity 2  window=2026-08-13T20:04:33Z  signer=0x82F2Dd9a…   complete
   entity 3  window=0 (UNSET)             signer=0x82F2Dd9a…   PARTIAL
   entity 4  window=0                     signer=0x0           never installed
   ```

2. **Teardown becomes impossible.** Revoking entity 3 hands back an `onChainCleanup` payload. Sending it from the owner EOA reverts at gas estimation:

   ```
   execution reverted (unknown custom error), data: 0xa24a13a6
   0xa24a13a6 = ArrayLengthMismatch()
   ```

   The payload carries three teardown entries; the account has two hooks.

3. **Every later grant fails.** A grant's deferred install is an install/**replace** batch — tear down the priors, install the new entity. The teardown reverts, so the batch reverts:

   ```
   SESSION_GRANT_INSTALL_FAILED: deferred grant install/replace did not land
   (new entity not installed; prior entities not torn down):
   gas manager declined to sponsor: ... validation reverted: [reason]: AA23 reverted
   ```

   Entity 4 has been stuck `pending` ever since. The runner is bricked for new grants and cannot be repaired by re-granting, which is the only remedy the product offers.

## Mechanism

[...]

PR #768, opened 2026-08-23
fix: authenticate the operator gRPC plane, honoring declared alias keys

chrisli30:
## Summary

The `aggregator.Node` gRPC service — the operator plane — had no authentication. Every RPC on it either mutates task state or streams task data, and all of them were reachable by any caller that could open a socket to the port.

Two independent gaps produced this:

1. The service was registered on a bare `grpc.NewServer()` with no interceptors, so nothing sat between the wire and the handlers.
2. The one auth helper that existed, `verifyOperator`, short-circuited on a `const enforceAuth = false` left over from the 1.3 operator rollout — so even the single handler that called it (`Ping`) was not actually checking anything.

This came in as a community security report against `ReportEventOverload`, which calls `engine.DisableWorkflow` with no ownership or identity check. That RPC is real but it is the least interesting consequence; the root cause covers the whole service.

## Why this is worse than the reported RPC

The operator address was a self-declared field on the request. Nothing proved the caller held the corresponding key, which means the approved-operator allowlist was decorative — `CanStreamCheck` gates task distribution on an address the caller simply asserts.

- `SyncMessages` streams `TaskMetadata` (task id + full trigger config) for every active task to any caller claiming an approved operator address.
- `n.operatorStreams[address] = srv` is keyed by that same claimed address, so connecting as an existing operator evicts the real one from the notification path.
- `NotifyTriggers` reaches `QueueExecutionData`, i.e. it can force real workflow executions.
- `ReportEventOverload` disables any workflow whose id you know — and step one hands you every id.

Task ids are ULIDs and not enumerable, which is the only reason the reported RPC was not trivially mass-exploitable on its own. That mitigation disappears once `SyncMessages` is in the chain.

## The fix

Enforcement moved into unary and stream interceptors attached at `grpc.NewServer()` construction, in `aggregator/operator_auth_interceptor.go`. Two properties motivated putting it there rather than in each handler:

- There is no window in which the service is registered without the check.
- A handler added later inherits it instead of having to remember it. A per-handler check is one forgotten call away from an unauthenticated state change, which is precisely how `ReportEventOverload` shipped.

The model is **default-deny**. A method is authenticated unless it is listed in `publicNodeMethods`, and an authenticated method must carry an operator address for the signature to bind to. A request type with neither is refused rather than served anonymously — so a future RPC cannot silently inherit anonymous access. `HealthCheck` is the only public method: it reads and writes nothing, and operators call it before they have working credentials to tell "aggregator is down" from "my key is wrong".

Streaming RPCs carry their operator address in the first client message, which has not arrived when the interceptor runs. The stream is wrapped so the check happens the moment that message is received and still before the handler is given it.

Every failure returns the same opaque `Unauthenticated`. Reasons are logged, never returned — an unauthenticated caller should not learn whether an address is known, whether a signature parsed, or which of the two it got wrong.

## Operator alias keys — why the obvious fix was wrong

The first pass compared the recovered signature against the request's `operator_address` and nothing else. That is wrong for this AVS, and it would have taken down most of the fleet.

Operators are not required to run with the key behind their registered EigenLayer address. They may keep that key cold and run the node with an **alias key**, declared against the registered address in the APConfig contract (`declareAlias` / `getAlias`). `Operator.Start` already validates this binding on the operator side, and the struct comment at `operator.go` says as much: *"signerAddress match operatorAddr unless the operator use alias key"*.

Checking mainnet APConfig for the three currently approved operators: one signs with its own key, **two sign with declared alias keys**. Comparing against `operator_address` alone would have refused both at their first heartbeat.

So `verifyOperator` now recovers the signer and accepts it as either the operator's own key or the alias it declared on chain:

- Alias lookups are cached for 5 minutes, keeping a chain call off the per-RPC auth path.
- On an RPC failure the resolver falls back to the last known mapping. A stale mapping is still the one the operator declared — changing it takes a deliberate transaction — so serving it beats disconnecting the fleet over an endpoint blip.
- The signed message names the operator being acted for, so an alias signature cannot be replayed to act as a different operator.

The approved-operator allowlist and its hardcoded fallback are deliberately **unchanged**. Since `approved_operators` is intentionally unset, that fallback is the production allowlist, and removing it would have stopped task distribution entirely. Those addresses were never the weakness — the inability to prove a claim to them was. Now that the claim is verified, the list is sound.

## Also fixed, found while sweeping for the same pattern

**Debug endpoints were live in production.** `/_debug/sentry/message` and `/_debug/sentry/panic` were gated on `APP_ENV != "production"`, but `APP_ENV` is set in no deployment — not in this repo, not in infra — so the condition was always true. Confirmed against the production gateway non-invasively: an unknown path returns 404, both debug paths return 405, i.e. the routes are registered. `middleware.Recover()` contains the panic, so the impact is unauthenticated Sentry flooding and arbitrary message injection via `?msg=`, not a crash. Both now gate on the `environment:` config field, which production actually sets and which reads as not-development when unset. gRPC reflection is gated the same way.

**Operator signatures had no upper epoch bound.** The freshness check only bit once a timestamp was in the past, so an epoch set far in the future minted a credential that never expired. Now bounded on both sides, with a 60s allowance for operator clock skew.

**Shared APConfig addresses.** The deployment addresses moved into `core/chainio/apconfig` so the operator (declaring an alias) and the aggregator (verifying against one) cannot drift apart.

## What was checked and left alone

The REST surface was swept for the same class of bug and is in good shape: `ensurePermission` is centralized and fails closed on an unmapped operation, only three operations are deliberately public, session policies enforce `requireOwnedWallet`, and the gas-manager webhook uses constant-time comparison and fails closed throughout. The gRPC operator plane was the one surface that never received that treatment.

Flagged but out of scope here:

- Operator transport is plaintext (`insecure.NewCredentials()`), so an on-path attacker can replay a captured header for the signature TTL. The real fix is TLS, which needs infra work and coordinated operator config changes.
- The worker gRPC server (`0.0.0.0:50051`) is also unauthenticated, but it is read-only chain proxying on the private network. Worth confirming no worker service has a TCP proxy attached.
- `/operator` and `/telemetry` publicly list operator addresses. Low value now that spoofing is closed — they are on-chain anyway — but it is what hands an attacker the addresses to try.

## Deployment notes

- The aggregator now binds the APConfig contract at startup and **hard-fails if it cannot reach the AVS chain**. This is deliberate: a degraded resolver means refusing most of the fleet, so failing loudly beats starting into a silent outage. It does make the AVS-chain RPC a new startup dependency for the gateway.
- No storage or protobuf changes, so no migration.
- Operators need no changes. Both connection paths already attach `ClientAuth` per-RPC credentials, and alias-key operators keep working.

## Testing

14 new tests in `aggregator/operator_auth_interceptor_test.go` covering unauthenticated refusal on every RPC, impersonation, expired and far-future signatures, stream spoofing, alias accepted, undeclared key rejected, no-alias-must-use-own-key, `HealthCheck` staying public, and default-deny on an unmapped request type.

These were mutation-tested: reinstating the `enforceAuth` bypass fails 6 of them, so they detect the actual vulnerability rather than just exercising the new code.

`go build`, `go vet`, and package tests pass across `aggregator/...`, `operator/`, `worker/`, `core/auth`, and `core/chainio`.

### DIFF
--- a/aggregator/auth.go
+++ b/aggregator/auth.go
@@ -3,6 +3,9 @@ package aggregator
 import (
 	"context"
 	"fmt"
+	"strings"
+
+	"github.com/ethereum/go-ethereum/common"
 
 	"github.com/AvaProtocol/EigenLayer-AVS/core/auth"
 	"google.golang.org/grpc/metadata"
@@ -36,5 +39,33 @@ func (r *RpcServer) verifyOperator(ctx context.Context, operatorAddr string) (bo
 		return false, fmt.Errorf("missing auth header")
 	}
 
-	return auth.VerifyOperator(authRawHeaders[0], operatorAddr)
+	signerAddress, err := auth.OperatorSignerFromAuthHeader(authRawHeaders[0], operatorAddr)
+	if err != nil {
+		return false, err
+	}
+
+	// The common case: the operator signed with its own registered key.
+	if strings.EqualFold(signerAddress.Hex(), operatorAddr) {
+		return true, nil
+	}
+
+	// Otherwise the signature only counts if it came from the alias key
+	// this operator declared on chain. Resolving that mapping is what
+	// makes an alias-key operator distinguishable from an impostor.
+	if r.aliasResolver == nil {
+		return false, fmt.Errorf("signature is not from operator %s and alias resolution is unavailable", operatorAddr)
+	}
+
+	alias, err := r.aliasResolver.aliasFor(ctx, common.HexToAddress(operatorAddr))
+	if err != nil {
+		return false, fmt.Errorf("resolving alias for operator %s: %w", operatorAddr, err)
+	}
+	if alias == (common.Address{}) {
+		return false, fmt.Errorf("operator %s declared no alias and did not sign with its own key", operatorAddr)
+	}
+	if alias != signerAddress {
+		return false, fmt.Errorf("signature is from neither operator %s nor its declared alias", operatorAddr)
+	}
+
+	return true, nil
 }
--- a/aggregator/auth.go
+++ b/aggregator/auth.go
@@ -14,17 +14,18 @@ import (
 // helper they shared with the other Aggregator handlers is no longer
 // needed because REST middleware verifies JWTs in the handler stack.
 //
-// What stays in this file: verifyOperator, the operator-stream gRPC
-// auth check used by the Node service. Operators continue to speak
-// gRPC.
+// What stays in this file: verifyOperator, the operator gRPC auth check
+// used by the Node service. Operators continue to speak gRPC. It is
+// applied to every Node RPC by the interceptors in
+// operator_auth_interceptor.go, not by individual handlers.
 
-// verifyOperator checks validity of the signature submit by operator related request
+// verifyOperator checks validity of the signature submit by operator related request.
+//
+// Callers get (false, err) on every failure; there is no configuration
+// that turns this into a pass. It used to short-circuit on an
+// `enforceAuth = false` constant left over from the 1.3 operator
+// rollout, which made the whole Node service callable by anyone.
 func (r *RpcServer) verifyOperator(ctx context.Context, operatorAddr string) (bool, error) {
-	// TODO: Temporary not enforce auth
-	if !enforceAuth {
-		return true, nil
-	}
-
 	md, ok := metadata.FromIncomingContext(ctx)
 	if !ok {
 		return false, fmt.Errorf("cannot read metadata from request")
@@ -37,7 +38,3 @@ func (r *RpcServer) verifyOperator(ctx context.Context, operatorAddr string) (bo
 
 	return auth.VerifyOperator(authRawHeaders[0], operatorAddr)
 }
-
-// enforceAuth toggles operator-side auth enforcement. Same default as
-// the legacy file: false until all operators upgrade past 1.3.
-const enforceAuth = false
--- a/aggregator/health_deep.go
+++ b/aggregator/health_deep.go
@@ -69,12 +69,24 @@ type operatorHealth struct {
 	Addresses []string `json:"addresses"`
 }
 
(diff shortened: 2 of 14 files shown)


chrisli30:
## Audit of the alias follow-up (`8959ad26`) plus one deploy-breaking gap

The alias model is correct: `operator_address` is the EigenLayer identity, the keystore is allowed to be a declared hot key, and the signed message names the operator so an alias cannot impersonate another. The boot-time equality check is gone. Cache + fail-open on RPC blip is the right trade.

**What would still have taken down the fleet:** production `gateway-railway.yaml` has `eth_rpc_url: ${SEPOLIA_RPC}` (with a comment to "flip to Ethereum when enforceAuth is on"). Alias lookup was bound only to that RPC, so `AddressForChain` picked **testnet** APConfig. The two alias-key operators (`0xc6b87cc9…`, `0xa026265a…`) declared on **mainnet** APConfig. Sepolia `GetAlias` would return zero → "declared no alias" → refused at first heartbeat. The Sepolia operator (`0x997e5d40…`, no alias) would have been fine.

`a800b5c0` binds both: Sepolia from `eth_rpc_url`, Ethereum from `chains[].ethereum` when present. Startup `CodeAt`s each APConfig (the actual new dependency — Dial alone is lazy). `/health/deep` now reports `apconfig[]`. A failed lookup on one chain is not cached as "no alias".

**Pre-deploy check:** `ETHEREUM_RPC` on the gateway service must answer `eth_getCode` for `0x9c02dfc9…` (mainnet APConfig). If it does not, the gateway will refuse to start. Do not flip top-level `eth_rpc_url` to Ethereum — that would break the Sepolia operator. Both APConfigs are consulted.

**Left alone:** plaintext operator transport, worker gRPC, public `/operator` listing, #766 teardown. Disclosure: this PR is already public; no extra write-up from me.

PR #788, opened 2026-09-17
docs: plan native ETH session permission and EOA 7702 grants

chrisli30:
## Summary

Proposed spec for two independently shippable permission tracks. **No runtime code in this PR.**

- **Track A** — Native ETH send (`ethTransfer` / withdraw) on the derived MA v2 smart wallet. Today every REST grant is selector-scoped, so empty-calldata `execute(to, value, 0x)` is blanket-refused (`SESSION_POLICY_NATIVE_NOT_ALLOWED`).
- **Track B** — Same permission vocabulary on the user's EOA via Alchemy `SemiModularAccount7702`. Calibur is not used. Spike first; production execute stays flag-off.

**K0 (purpose-matched grants):** compile every permission the stated intent needs. A Uniswap consent includes router + cap token + WETH approve/deposit/withdraw so ETH-in and ETH→WETH demote work. Uniswap does not need a native-send grant. Native-send is a separate purpose, merged into the same singleton grant only when the workflow also sends ETH.

## Why this is a docs PR

Implementation is split (A0–A5, B0–B6) in the spec. This lands the plan under `docs/changes/` so it is tracked the same way as other architecture notes.

## Test plan

- [x] Spec-only; no Go/OpenAPI/storage change
- [ ] Review K0–K13 and the PR plan in the doc before any A0/B0 spike

See `docs/changes/20260917-native-eth-and-eoa-permissions.md` for the full plan.

Related: discussion #658, `PLAN_PARTNER_PAYMENTS.md` §4.1, `FINDINGS_AA23_WETH_SELL_SESSION_SCOPE.md`, `docs/changes/20260806-session-grant-replace-on-submit.md`.


chrisli30:
Folded in: multi-token ERC-20 caps are **not** Track A and **not** A1.

- Alchemy AllowlistModule already caps **per token**. Our wire is still one `{ token, amount }`. Allowlisting WETH without a WETH cap leaves WETH uncapped (#1674).
- Native ETH is **not** that field — Track A `nativeSpendCap` + `NativeTokenLimitModule`. There is no dedicated ETH grant on `erc20SpendCap` today.
- A second Send ERC-20 catalog row cannot add a token (one usable grant per chain+runner).
- **A7** (after A3): additive `erc20SpendCaps[]`, pack `HasERC20SpendLimit` per token, Studio amount list on the same grant. See K15 / §A.7.

A0–A5 stay native-send only.

chrisli30:
Correction: **A7 (`erc20SpendCaps[]`) is UI-blocking**, not a post–Track A nice-to-have.

Studio cannot honestly ship the shared-grant amount list (Uniswap + WETH cap, merge without dropping a token) on a single `{ token, amount }`. A7 does **not** need native ETH — it can start **now**, in parallel with A0. It must **not** wait for A3.

Native send remains A0–A5. ETH still does not go in the ERC-20 cap list.

PR #790, opened 2026-09-18
test: spike native ETH session hooks on Sepolia and Base

chrisli30:
## Summary

Track A **A0** from `docs/changes/20260917-native-eth-and-eoa-permissions.md`. Live spike, not REST packing.

`go run ./scripts/spike/native_eth_hooks`

**Env:** `SPIKE_OWNER_KEY`, `SPIKE_CONTROLLER_KEY`, `SPIKE_BUNDLER_URL`, optional `SPIKE_RPC_URL` (default Sepolia publicnode), `SPIKE_SALT` (default 17), `SPIKE_SKIP_WIDE=1` to skip the 20-recipient probe.

**Proofs:** 1–7, 9 always; 8 gas table (20-row is A2-blocking — re-run without skip). Run twice: Sepolia and Base.

NT hook packing is **in the spike** (promote in A2). Uninstall uses **val-then-exec**, not flat reverse. After teardown it reads `NativeTokenLimitModule.limits` (receipt is not evidence).

Self-funded; no Gas Manager.

## Live evidence

Not run in this PR (no `SPIKE_*` in the agent environment). Paste `PROOF n PASS` lines + `GAS` table from Sepolia and Base here before merging A2.

## Test plan

- [x] `CGO_ENABLED=0 go build ./scripts/spike/native_eth_hooks`
- [ ] Sepolia full run (proofs 1–9)
- [ ] Base full run
- [ ] 20-row gas numbers recorded (or FINDING + recommended max/per-row seed)

PR #791, opened 2026-09-18
feat: pack NativeTokenLimitModule and val-then-exec uninstall

chrisli30:
## Summary

A2 packing from the native-ETH plan, after A0 Sepolia (#790).

- `NativeTokenLimitModule` at `0x00000000000001e541f0D090868FBe24b59Fbe06` (v2.0.0, not redeployed in v2.0.1)
- `PackNativeTokenLimitInstallData` / `UninstallData`, val + exec hook builders
- **Val-then-exec uninstall** (K10). 3-hook ERC-20 grants become `[TR-val, AL-val, empty]` (not `[TR, empty, AL]`). 5-hook mixed: `[TR, NT-entityId, AL, empty, empty]`
- NT teardown is `abi.encode(uint32 entityId)`, packed from NT's own uint32 ABI (not via `PackSingleSignerUninstallData`)
- VGL: `AllowlistRows` set in `NewSessionResolver` from packed install (same site as `DeferredTeardownCount`); +45k per row past 3 (two-point interpolation: 3 rows ≈ 700k, 20 rows ≈ 1.5M)

Does **not** wire REST native fields (A1) or preflight (A3). ERC-20-only `HooksFor` is unchanged.

A2 **does not** use the lucky 1.8M replace VGL; deferred replace AA23'd 3/3 at 2.2M. Uninstall exec cost from A0: `eth_estimateGas` 448k (owner-tx, not UserOp VGL) — keep `seedVerificationGasPerUninstall` at 120k. Product max native recipients is **5** (`aa.MaxNativeRecipients`): 20-row install passed, 20-row replace does not.

## Review follow-up (`3b13bfc1`)

- **F1:** `CountAllowlistInputs` + `auth.AllowlistRows` in production
- **F2:** leave 120k; 448k is 20-row owner-tx estimate, moot at max 5
- **F3:** spec K3/K14/Q3 + `aa.MaxNativeRecipients = 5` for A1
- **F4:** dedicated NT uninstall packer via `uint32OnlyArgs`

## Test plan

- [x] `go test ./core/chainio/aa/ ./pkg/erc4337/preset/`
- [x] `go test ./core/taskengine/ -run SessionGrant|SessionPolicy|Teardown|Uninstall|HooksFor|SessionResolver`


chrisli30:
Review follow-up in `3b13bfc1`:

**F1** — `AllowlistRows` is set in `NewSessionResolver` next to `DeferredTeardownCount`, from packed AllowlistModule inputs (`aa.CountAllowlistInputs`), not `len(AllowedActions)`. The 45k/row K14 scale now fires on the production send path.

**F2** — left `seedVerificationGasPerUninstall` at 120k. The 448,481 figure is owner-tx `eth_estimateGas` of a 20-row 5-hook uninstall, not UserOp VGL. Raising it does not save 20-row replace (AA23 3/3 at 2.2M).

**F3** — product max is 5 (`aa.MaxNativeRecipients`). Spec K3/K14/Q3/A1 updated. 20-row *install* passed at 1.5M; 20-row *replace* is the singleton re-grant path and does not. A1 must refuse `len(nativeRecipients) > 5`.

**F4** — `PackNativeTokenLimitUninstallData` packs from `uint32OnlyArgs` (NT's own ABI). SingleSigner/TimeRange uninstalls use the same primitive, each with its own function.

Nits: SSTOREs; 45k comment is the two-point interpolation; stray `EOF` stripped from this body.

chrisli30:
N1 in `99bfdcb2`.

`CountAllowlistInputs` now returns `0, nil` on any decode/unpack miss (including a ≥29-byte malformed install). `newSessionResolver` logs and falls back to 0 instead of failing the send. Zero is the field's documented unknown → 700k seed. Under-seeding is AA26 at estimation.

The `len(inner) < hookConfigLen+4` stub special-case is gone; tests and predating grants take the same path.

Minors: `TestMaxNativeRecipientsFollowsReplaceNotInstall` notes nothing enforces the cap until A1. Left `MaxNativeRecipients` in `aa` for this PR — A1 Validate is the natural move to `taskengine`. Left `reflect` for ABI-decoder shape changes.

chrisli30:
Writeup precision, in `7b37a52f`.

Fail-open restores **sends** only for grants at or under 3 allowlist rows (0–3 are the same 700k seed). An undecodable install that actually carries more than 3 converts “resolver errors, every send fails” into “700k seed, first-op AA26 at estimation” — better failure mode, not a working send, not a fund risk.

Who that leaves unrescued: a **4+ target grant whose InstallCall we cannot decode**. Today that is a custom / merged grant with extra allowed-action targets (A7 extra cap tokens on top of Uniswap’s typical router + cap token + WETH = 3). Those grants already ran on the unwired 700k seed before `AllowlistRows` was populated; a decode miss puts them back there. Typical Uniswap (≤3) is rescued. A1 native recipients are new rows and will decode from current packing.

Also in this commit:

- `TestSeedVerificationGasScalesWithGrantContents` now pins `AllowlistRows: 0` and `: 3` to `seedVerificationGasDeferredHooks` (700k), so the resolver’s fail-open and the seed math are one assertion, not a reader trace.
- Decode misses return `0, nil`, so the guessed-seed **warn** is now on `rows == 0 && RequiresExecuteUserOp`, not on `rowErr`. The resolver test spies the logger. `logGuessedAllowlistSeed` is a no-op when `globalLogger` is unset (tests); `Engine.SetLogger` runs before the production send path.

PR #792, opened 2026-09-18
feat: add native session-grant fields to OpenAPI and storage model

chrisli30:
## Summary

A1 from the native-ETH plan. A7 and A2 are already on staging, so this both adds the REST/storage fields **and** packs them (the original A1 “not implemented” gate is obsolete).

- OpenAPI: optional `nativeRecipients` (max 5), `nativeSpendCap`, `allowContractRecipient`. `allowedActions` / `erc20SpendCap` leave `required`. Present `[]` is 400.
- Breaking generated Go: those fields are pointers; `permissionsFromAPI` nil-guards them.
- `SessionPermissions.Validate`: native-only, mixed, nativeValueCap; reserved modules/factory/EntryPoint; overlap with `allowedActions` refused; `eth_getCode==0` unless the flag (fail-closed with no reader).
- `HooksFor`: ERC-20-only stays `[AL-val, AL-exec, TR]`. Native spend cap adds NT val+exec. Native recipient rows set `HasSelectorAllowlist=false`.
- `MaxNativeRecipients = 5` lives next to Validate (A0: 20-row replace AA23s).
- Storage additive (`make storage-check` vs staging green). Display fields only; signed truth remains `Grant.InstallCall`.

Does **not** change execution preflight (A3). Empty-calldata `ethTransfer` is still refused on MA v2 until A3 reads `nativeRecipients`.

## Test plan

- [x] `go test ./aggregator/rest/ ./core/taskengine/ ./core/chainio/aa/`
- [x] `go run scripts/compare_storage_structure.go origin/staging`


chrisli30:
Review follow-up in `588c39e2`.

**F1** — `HooksFor` now runs static `validate(false)` so it does not repeat `eth_getCode`. Production `CodeAt` uses `GetChainStateReaderForChain` (pooled) and `memoizeCodeAt` so a 5-recipient grant is 5 lookups on one reader, not 20 dials. `TestHooksForDoesNotRepeatCodeAt` and `TestPrepareSessionPolicyLooksUpEachRecipientOnce` pin it.

**F2** — dropped `aa.EntrypointAddress` from `reservedNativeRecipient`. v0.7 `config.EntryPointV07AddressHex` stays as the constant.

**F3** — nil `CodeAt` names `InstallSessionResolver`.

**F4** — `logContractRecipientException` after a successful validate, and only when `len(NativeRecipients) > 0`.

**F5** — comments on `MaxNativeRecipients` and OpenAPI `maxItems: 5` point at each other.

**Controller** — `SessionSigner` is bound at prepare/submit and refused as a native recipient (`TestPrepareSessionPolicyRefusesControllerAsNativeRecipient`).

chrisli30:
Residual hardening in `dca53b2d`.

**HooksFor** requires a non-nil `CodeAt` whenever `NativeRecipients` is set and `AllowContractRecipient` is false. Nil-check only — zero lookups — so `TestHooksForDoesNotRepeatCodeAt` still passes at 1, and an unbound caller (Track B packing without `Validate()`) fail-closes instead of emitting unscoped rows. `TestHooksForRefusesUnprovenNativeRecipients` pins it.

Nits:

- `memoizeCodeAt` comments that the cache is unsynchronized and assumes single-goroutine use per request.
- `GetChainStateReaderForChain` is resolved inside the lookup, not captured at bind.
- Missing controller key is now an error from `bindNativeRecipientChecks`, not a skipped reservation.

